PR #30, opened 2015-02-05
MS SQL support for data importer

shoe54:
Uses jtds JDBC library which supports both MSSQL and Sybase. Note that jtds license is LGPL which may be incompatible with Apache License, but I noticed that mysql connector is used (GPL) and may be similarly incompatible.


prasser:
Thanks for your interest in ARX and for your contribution!

You are right regarding the license incompatibility. We are currently working on a solution for user-initiated dynamic loading of JDBC drivers to overcome licensing issues in the future. 

For now, I will merge your pull request so that ARX provides in-tree support for MS SQL. In a future release, we will externalize jtds together with the mysql connector.

PR #61, opened 2016-03-16
Improve entropy-l-diversity criteria class

sebastianst:
The entropy-l-diversity checking method `isAnonymous()` is improved in several ways:

1. Omit log-base 2 normalization, since the underlying inequality is linear in the logarithm and can thus be written in any base.
2. Use `entry.count` instead of creating the `total` sum during the for loop.
3. The underlying inequality can be written so that the sum must not be larger than the introduced constant `C`. This allows to abort the for loop earlier, as soon as the sum grows over `C`.


prasser:
Thx. Merged!

PR #68, opened 2016-08-11
Fix how outlying tuples are counted

mbode:


prasser:
Dear Maximilian,

thanks for your interest in ARX!

I have checked the pull request and the associated issue. As for as I can tell, this is not a bug, because by definition the TupleWrapper will always collapse all suppressed records into a single group (i.e. equivalence class). While I agree that the code is not very intuitive, I think the assigment "numberOfOutlyingTuples = element.getCount()" is correct. If you disagree, please provide us with an example that surfaces a bug.

Best regards
Fabian

P.S.: I would happily accept a pull request which documents why the assignment seems to be so weird...

mbode:
Hi Fabian,

thank you for responding so quickly. I see, that makes sense.
We tried updating from 3.3.0 to 3.4.1 and saw some changes in statistics, it seemed like outliers where not counted at all. (All of `numberOfEquivalenceClasses`, `maximalEquivalenceClassSizeIncludingOutliers` and `numberOfOutlyingTuples` were different.) Maybe there was a change in how suppression is handled – can you shed some light on this?

Best regards,
Max

prasser:
Hi Max,

I just compared the two version of the code and I can confirm that there were some changes on how suppressed records are determined:

In version 3.3.0. a record was considered to be suppressed when the values of all quasi-identifiers were identical to the string used for representing suppressed values ("*" in ARX's default configuration). In  version 3.4.1. we only consider records as suppressed which have actively been suppressed by the system. This means, e.g., that a record which looks as if it was suppressed because all values have been generalized to "*", will *not* be considered suppressed.

There are valid arguments for both ways to calculate the statistics.

It would be great if you could analyze your results in more detail to find out whether you discovered a bug or whether the observed behaviour is as expected. For example, are you sure that only the statistics have changed while the data which is described by the statistics has not changed? Can the changes observed by you be explained with my description of the changes in ARX, i.e. do you calculate statistics for datasets in which records have been "suppressed" by generalization only? Do you use local recoding? Does the change which you propose in this PR change the behaviour of ARX in your case (in theory, it should not change the behaviour of the system)?

Thanks and regards
Fabian

P.S.: If this does not turn out to be a bug, would you be willing to provide a PR which documents the strange line of code?

prasser:
Dear Maximilian,

as you are not responding, I have added a code comment myself and I will close this PR now. If your problem persists, please open a new issue.

Thanks
Fabian

PR #72, opened 2016-10-04
Grassberger l diversity

sebastianst:
Add possibility to choose the entropy estimator (Shannon or [Grassberger](https://arxiv.org/abs/physics/0307138v2)) during the instantiation of an `EntropyLDiversity` criterion.
I realized this by adding an enumerator `EntropyEstimator` to the `EntropyLDiversity` class, which basically selects the logarithm subsitute in the entropy estimation formula.

The default is, as was the only option before, the naive Shannon estimator. This also reflects the original definition of entropy-l-diversity. To keep the class compatible with old serializations, I added a custom deserializer, which first executes the default deserialization and then sets the estimator to Shannon if it is `null` after deserialization. This custom deserialization has not been tested yet.

The newly added option is to use the Grassberger estimator, see [this paper](https://link.springer.com/chapter/10.1007/978-3-319-45381-1_11) for a comparison to the default Shannon estimator.

The unit test `TestAnonymizationLDiversity` completes successfully. **It remains to add test cases for the Grassberger entropy estimator.**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 /junit_test_hierarchy_age.csv
 /jars
 /test.ahs
+/build
--- a/src/main/org/deidentifier/arx/criteria/EntropyLDiversity.java
+++ b/src/main/org/deidentifier/arx/criteria/EntropyLDiversity.java
@@ -17,6 +17,9 @@
 
 package org.deidentifier.arx.criteria;
 
+import java.io.IOException;
+import java.io.ObjectInputStream;
+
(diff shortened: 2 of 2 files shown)


sebastianst:
BTW I used lambdas. In case you require 1.7 compatibility, the lambdas in the enum definition must be changed to anonymous classes implementing `IPsi`.

prasser:
Thanks for your contribution! I have just merged the code.

PR #81, opened 2016-11-03
Monotonic with generalization flag added to utility measures

jgaupp:
Numerous minor updates as I added the monotonic with generalization flag to constructors throughout the metric class hierarchy, rather than relying a default in the super class.  I figured it was important for documentation and readability to explicitly distinguish suppression and generalization when instantiating.

Tested by comparing the result lattices from equivalent anonymization ops in 3.4x and 3.5, using a mix of models and utility measures.


src/main/org/deidentifier/arx/ARXLattice.java
```diff
@@ -1265,6 +1263,13 @@ private void readObject(ObjectInputStream aInputStream) throws ClassNotFoundExce
         }
     }
     
+    private void setMonotonicity(boolean isSuppressionAlwaysEnabled, int absoluteMaxOutliers) {
```
jgaupp: When testing ARXConfiguration changes, I knew what to look for and have general confidence in the effect of the changes.  I'm not as confident in the ARXLattice changes.  Seems like these evaluations drive a specific optimization of the search process, replacing utility IL bounds with node IL bounds.  I tested as no change from baseline but haven't configured a test of the "new code path" that would be used by utilities with non-monotonic generalization.

src/main/org/deidentifier/arx/metric/Metric.java
```diff
@@ -1143,8 +1143,11 @@ protected static MetricDescription getDescription(Metric<?> metric) {
 
     /** Is the metric independent?. */
     private boolean           independent      = false;
+    
+    /** Is the metric monotonic with generalization?. */
+    private boolean           monotonicWithGeneralization    = true;
 
-    /** Is the metric monotonic?. */
+    /** Is the metric monotonic with suppression?. */
     private boolean           monotonic        = false;
```
jgaupp: Per recommendations, retained field name for backwards compatibility in serialization.  All other changes are cool, I think.

src/main/org/deidentifier/arx/metric/MetricAECS.java
```diff
@@ -46,7 +46,7 @@
      * Creates a new instance
      */
     protected MetricAECS() {
-        super(false, false);
+        super(true, false, false);
```
jgaupp: All utilities, except publisher's payout, have monotonic with generalization set to true.  Super class default is also true.

src/main/org/deidentifier/arx/metric/MetricDefault.java
```diff
@@ -1,77 +1,77 @@
-/*
```
jgaupp: Not sure why the review identifies some classes as completely replaced.  No wholesale cut and paste on my part, though I did rely on automated refactoring.

src/main/org/deidentifier/arx/ARXConfiguration.java
```diff
@@ -707,7 +707,9 @@ public Monotonicity getMonotonicityOfPrivacy() {
      * @return
      */
     public Monotonicity getMonotonicityOfUtility() {
-        if (metric.isMonotonic() || (this.getMaxOutliers() == 0d) || this.isPracticalMonotonicity()) {
+        if ((metric.isMonotonicWithGeneralization() && this.getMaxOutliers() == 0) ||
```
jgaupp: The action code is here (plus equivalent in lattice and lightning).  My understanding is that generalization and suppression are mutually exclusive when associated with number of outliers, so all evaluations are setup as ((&&) || (&&)).

In my first change, I actually bungled this logic and testing by comparison of 3.4 to 3.5 failed.  I was a bit surprised that unit tests did not reflect the issue (all passed).  Would tests for a specific generalization scheme be in order?  Seems like the existing utility tests are based on information loss?

prasser: It would probably be a good idea to take a look at this in more detail. How did the "bungled logic" look like? Which comparison failed but was not reflected by the unit tests? Can you provide me with code?

src/main/org/deidentifier/arx/ARXLattice.java
prasser: I'm also not sure that the changes in ARXLattice are correct. The affected logic is pretty complex and I'm having a hard time remembering exactly how this works... :)

It is not used for optimizing the search process (the ARXLattice is always constructed after the search process) but for estimating the data quality/utility/info.-loss of transformations which have been pruned during the search process (the estimation is implemented in class UtilityEstimator, which also contains a (short) description of the process). I will try to wrap my head around it.

src/main/org/deidentifier/arx/ARXConfiguration.java
jgaupp: Consider this a false alarm.  I went back through my local history this morning, tested out my early code that didn't tie generalization to outliers, and tested a purposeful mismatch of outliers and monotonicity types.  Unit tests caught both cases.  I think I was already worried about results and convinced myself something was amiss.

prasser: Ok, great. I think that this bug is fixed. However, we don't have a unit test for the Stackelberg game that benefits from the new code path (assuming monotonicity without it actually being present does not necessarily lead to wrong results and this fix only affects configurations with a suppression limit of 0%). Have you checked that ARX is now executing the correct algorithm?

PR #89, opened 2017-04-03
Bug: NPE when saving project without data

eicherj:
Bug: NPE when saving project without data, leads to InfoDialog with empty text. Project-file is created in file system regardless.
Fix: If input data is null when saving project, show according error message, cancel saving action.


prasser:
I don't think that this is the correct fix for the problem. Why shouldn't users be able to save projects without data? They could, e.g., have specified a name and description for the project, changed some project settings. The correct fix would be to investigate and solve the NPE...

PR #93, opened 2017-04-12
Apply existing order of attribute values in the attribute order selection dialog

silvercluster:
(1) Apply existing order of attribute values in the attribute order selection dialog (if existing).
(2) Right align button bar in the attribute order selection dialog
(3) Rename buttons in the attribute order selection dialog and add button
names to resources (Load -> Load from file, Save -> Save to file,
OK -> Apply)
(4) Rename "Default" ordering label to "Custom"


prasser:
This PR has way too many changes and mixes unrelated stuff. Please separate (1) from (2)-(3) into different PRs.

PR #94, opened 2017-04-13
Apply an existing ordering to the attribute values when the order selection dialog for an attribute is opened.

silvercluster:
Issue description:
When the an attribute already has a stored order, the existing attribute
value ordering is not displayed when the dialog to specify the attribute
value ordering is reopened. Instead the distinct values are shown in the
occurrence order from the input data.

Fix 1:
It applies the existing attribute value order for the attribute values,
if existing. New attribute values that are not present in the existing
order of the attribute are appended to the end of the attribute value
array.

Fix 2:
In case an existing order is loaded from a file a check is included,
when the dialog is closed, that removes entries that are not present in
the input data set.

Fix 3:
When the "DialogOrderSelection" is canceled via the cancel button or the
close icon, the attribute type remains unchanged.


silvercluster:
Too many changes in this PR. Changes will split to several pull requests.

PR #96, opened 2017-04-13
Leave data type unchanged when cancel is pressed in the dialog

silvercluster:
When "attribute value order selection dialog" aborted via cancel button or closed (dialog close icon) then the data type of the attribute remains unchanged.


silvercluster:
Lines 150, 151: I would not change this as this is an error handling in case the determined order is invalid
Lines 164, 171: I agree, the change shall be done analogously.

I commited the relevant changes.

PR #97, opened 2017-04-13
Validate file import for "attribute value order selection dialog"

martinwaltl:
It is validated if the selected file contains the values of the
attribute's domain. If the file contains no or too few valid values a
warning dialog is displayed. If the file contains more values these
values are ignored and an info dialog is shown to the user.
Therefore, a warning dialog from JFace has been added to the controller.


silvercluster:
Simplified: Only files that contain the exact values present in the attributes' domain can be imported. In case the file contains no values, too few or too many values no values are imported.

The warning dialog is removed and the message are notified via info dialogs to the user.

src/gui/org/deidentifier/arx/gui/resources/messages.properties
```diff
@@ -936,6 +936,10 @@ DialogOrderSelection.12=Import...
 DialogOrderSelection.13=Export... 
 DialogOrderSelection.14=OK 
 DialogOrderSelection.15=Cancel 
+DialogOrderSelection.16=Import failed
+DialogOrderSelection.17=The file does not contain the values of the attribute's domain. Please select a file that contains only the values of the attribute's domain.
+DialogOrderSelection.18=Import failed
+DialogOrderSelection.19=The file contains more values than present in the attribute's domain. Please select a file that contains only the values of the attribute's domain.
```
prasser: Please simplify the messages
(1) Remove 16 and 18 "Import failed". I think that somewhere in the file, there is a generic "Error" title.
(2) One error message is enough: "The values contained in the file do not match the values of the attribute".

silvercluster: (1) This would be possible, but would result in an inconsistency as DialogOrderSelection.10, 3, 5 and 8 are also titles from error messages. Therefore, I would opt for creating a new seperate title message.
(2) Done

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogOrderSelection.java
```diff
@@ -303,19 +303,42 @@ protected void createButtonsForButtonBar(final Composite parent) {
...
+
+                            // Set items
+                            elements = fileData.toArray(new String[fileData.size()]);
+                            list.setItems(elements);
                         }
-                        
-                        // Set items
-                        elements = array;
-                        list.setItems(array);
+                    } else {
```
prasser: Does "file==null" really indicate an empty file? I think it indicates that the user has clicked cancel in the file dialog. What does the error message say? Is it actually needed? Wouldn't it be OK to silently exit the method, if no file was selected?

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogOrderSelection.java
```diff
@@ -199,7 +199,7 @@ private boolean isValidDataType(DataType<?> type, String[] values) {
      * @param charset TODO
      * @return
      */
-    private String[] loadArray(String file, Charset charset) {
+    private ArrayList<String> loadFile(String file, Charset charset) {
```
prasser: Please use the abstract return type List<String>

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogOrderSelection.java
```diff
@@ -199,7 +199,7 @@ private boolean isValidDataType(DataType<?> type, String[] values) {
      * @param charset TODO
      * @return
      */
-    private String[] loadArray(String file, Charset charset) {
+    private ArrayList<String> loadFile(String file, Charset charset) {
         ArrayList<String> list = new ArrayList<String>();
```
prasser: Please use the abstract type List<String> for defining the variable

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogOrderSelection.java
```diff
@@ -303,19 +303,42 @@ protected void createButtonsForButtonBar(final Composite parent) {
...
                 if (file != null){
-                    String[] array = loadArray(file, Charset.defaultCharset());
-                    if (array != null) {
-                        
-                        // Select string
-                        for (int i=0; i<combo.getItems().length; i++){
-                            if (combo.getItem(i).equals("String")) { //$NON-NLS-1$
-                                combo.select(i);
+                    ArrayList<String> fileData = loadFile(file, Charset.defaultCharset());
+                    long fileDataLines = fileData.size();
```
prasser: Does size() really return a "long". It's likely an "int".

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogOrderSelection.java
```diff
@@ -303,19 +303,42 @@ protected void createButtonsForButtonBar(final Composite parent) {
...
                 String file = controller.actionShowOpenFileDialog(getShell(), "*.csv"); //$NON-NLS-1$
                 if (file != null){
-                    String[] array = loadArray(file, Charset.defaultCharset());
-                    if (array != null) {
-                        
-                        // Select string
-                        for (int i=0; i<combo.getItems().length; i++){
-                            if (combo.getItem(i).equals("String")) { //$NON-NLS-1$
-                                combo.select(i);
+                    ArrayList<String> fileData = loadFile(file, Charset.defaultCharset());
```
prasser: It seems to be possible that loadFile() returns "null". You need to either change "loadFile" or check for "null" here.

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogOrderSelection.java
```diff
@@ -303,19 +303,42 @@ protected void createButtonsForButtonBar(final Composite parent) {
...
+                    long fileDataLines = fileData.size();
+
+                    if (fileDataLines != 0) {
+                        // remove values that are not present in the attribute's domain
```
prasser: Please start comments with uppercase letters. Formulate complete sentences where appropriate.

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogOrderSelection.java
```diff
@@ -303,19 +303,42 @@ protected void createButtonsForButtonBar(final Composite parent) {
...
+                        fileData.retainAll(Arrays.asList(elements));
+
+                        if (fileData.isEmpty() || fileData.size() != elements.length) {
+                            // import failed, file contains no or not all of the
```
prasser: Please start comments with uppercase letters. Formulate complete sentences where appropriate.

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogOrderSelection.java
```diff
@@ -303,19 +303,42 @@ protected void createButtonsForButtonBar(final Composite parent) {
...
+                            // attribute's domain
+                            controller.actionShowInfoDialog(getShell(),
+                                                               Resources.getMessage("DialogOrderSelection.16"),
+                                                               Resources.getMessage("DialogOrderSelection.17"));
+                        } else if (fileDataLines != fileData.size()) {
+                            // import failed, the file contains more values as
```
prasser: Please start comments with uppercase letters. Formulate complete sentences where appropriate.

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogOrderSelection.java
```diff
@@ -303,19 +303,42 @@ protected void createButtonsForButtonBar(final Composite parent) {
...
+                        // remove values that are not present in the attribute's domain
+                        fileData.retainAll(Arrays.asList(elements));
+
+                        if (fileData.isEmpty() || fileData.size() != elements.length) {
+                            // import failed, file contains no or not all of the
+                            // attribute's domain
+                            controller.actionShowInfoDialog(getShell(),
+                                                               Resources.getMessage("DialogOrderSelection.16"),
+                                                               Resources.getMessage("DialogOrderSelection.17"));
+                        } else if (fileDataLines != fileData.size()) {
```
prasser: This does not solve the problem very wel. First, it is akward that this condition is checked *after* values have been retained. This is very difficult to read, because the value of fileDataLines has been set *before* values have been returned. This should be the first check which is performed before values are even retained. However, I have another suggestion. Please throw an Exception in the method loadFile(...) as soon as more than elements.size() lines have been read. This immediately stops the loading of data, when it becomes obvious that something is wrong. This check can then be removed.

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogOrderSelection.java
silvercluster: I explicitly used "ArrayList" as I use the functionality "retainAll" in the selection listener for the import button, which is optional for the List interface (see: https://docs.oracle.com/javase/8/docs/api/java/util/List.html#retainAll-java.util.Collection-)

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogOrderSelection.java
silvercluster: I will check for "null" value.

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogOrderSelection.java
prasser: Yes, please. Also, I think that the call to retainAll will not be needed anymore, after this patch has been revised.

silvercluster:
Comments have been incorporated. New version available.

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogOrderSelection.java
silvercluster: Included the value existance check during file loading. If a value is not present within the attribute's domain the file loading is aborted. As a result the _retainAll_ call is not needed anymore.

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogOrderSelection.java
```diff
@@ -193,34 +195,56 @@ private boolean isValidDataType(DataType<?> type, String[] values) {
...
+     * @param charset
+     *            TODO
      * @return
+     * @throws IllegalStateException
+     *             The file contains more or additional values (lines) than
+     *             present in the attribute's domain
      */
-    private String[] loadArray(String file, Charset charset) {
-        ArrayList<String> list = new ArrayList<String>();
+    private java.util.List<String> loadFile(String file,
```
prasser: Please import "java.util.List" and simply use "List" here and in other places where "java.util.List" is explicitly referenced.

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogOrderSelection.java
```diff
@@ -193,34 +195,56 @@ private boolean isValidDataType(DataType<?> type, String[] values) {
...
             String line = reader.readLine();
+            Set<String> _elements = new HashSet<String>();
+            _elements.addAll(Arrays.asList(elements));
+
             while (line != null) {
                 list.add(line);
+                if (list.size() > _elements.size() || !_elements.contains(line)) {
+                    // The file contains more or additional values (lines) than
+                    // present in the attribute's domain
+                    throw new IllegalStateException();
```
prasser: I think, it would be a better idea to show the error dialog here and then simply return "null", please also implement the check whether list.size() != elements.length also at the end of this message.

prasser: You can then remove the check from "widgetSelected"

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogOrderSelection.java
martinwaltl: As there is also the SWT widget "org.eclipse.swt.widgets.List" used, the class _List_ is not uniquely referencable. I will import _java.util.List_ and reference the SWT widget _List_ using the complete namespace, as it is only used once when creating the widget.

martinwaltl:
Applied requested changes. Please verify latest commit.

PR #99, opened 2017-04-14
Score

RaffaelBild:
Implementation of data-dependent differential privacy


prasser:
This seems to be a great patch set. But please break this up into smaller pull requests that can be reviewed independently. For example:
(1) General changes (e.g. to messages.properties, libraries) that are not directly related to the new feature
(2) Addition of the feature to the ARX Core/API
(3) Addition of the feature to the ARX GUI

Thanks!
Fabian

PR #100, opened 2017-04-19
Add tab 'Classification accuracy' in output side (analyzing view)

eicherj:
Issue: Currently, the feature and class selection for the classification
analysis in the 'Analyze/enhance utility' view is handled via the tab
'Classification accuracy' on the left-hand 'input side', and it is used
for both analysis of the input and output datasets.

Enhancement: In order to make it clearer to the user that this selection is being applied to both input and output datasets, I added the tab 'Classification accuracy' to the right-hand 'output side' with a hint message explaining this fact.


prasser:
I don't understand the intention of this change. How does it look like? Please add a screenshot...

eicherj:
image

prasser:
I see. The view can definitely be improved!

Idea: Why not display features on the left and classes on the right, i.e. in two separate views? It would also be great to add additional methods for selecting/deselecting attributes, e.g. "select all", "select none", "search". The view is basically unusable at the moment for datasets with many attributes...

PR #101, opened 2017-05-03
Apply existing order to values for ARXOrderedString attribute

martinwaltl:
For attributes of types "ARXOrderedString" the existing attribute
ordering is applied when the order selection dialog is opened.
For any other attribute types the values are sorted.


martinwaltl:
I assume your solution does not resolve the issue, as in the current version the _distinct_ values of the input data column are passed to the dialog and not the existing order of the column (stored in the DataType).
Nevertheless, I also see that my proposal contains unnecessary code, as I can always assume that the DataType contains always all elements from the input data set (as the DataType is updated when changed) and no additional rows are added unless a new data set is loaded.
I will simplify my proposal and commit my changes.

martinwaltl:
Please verify the new commit.

PR #113, opened 2017-07-03
Implement various score functions

RaffaelBild:


src/main/org/deidentifier/arx/metric/v2/MetricMDNUEntropyPrecomputed.java
```diff
@@ -333,12 +341,34 @@ protected void initializeInternal(final DataManager manager,
...
         }
         
+        if (config.isPrivacyModelSpecified(DataDependentEDDifferentialPrivacy.class)) {
+            
+            // Store minimal size of equivalence classes
+            DataDependentEDDifferentialPrivacy dpCriterion = config.getPrivacyModel(DataDependentEDDifferentialPrivacy.class);
+            k = (double)dpCriterion.getK();
+            
+            // Store root values of generalization hierarchies or null if no single root value exists
+            rootValues = new Integer[hierarchies.length];
```
prasser: Why is this array of type Integer[]. int[] is sufficient. You can encode a missing root value as -1, for example.

src/main/org/deidentifier/arx/metric/v2/MetricMDNUEntropyPrecomputed.java
```diff
@@ -326,7 +341,73 @@ protected void initializeInternal(final DataManager manager,
             max[i] = (input.getDataLength() * log2(input.getDataLength())) * Math.max(gFactor, sFactor);
         }
         
+        if (config.isPrivacyModelSpecified(DataDependentEDDifferentialPrivacy.class)) {
+            
+            // Store minimal size of equivalence classes
+            DataDependentEDDifferentialPrivacy dpCriterion = config.getPrivacyModel(DataDependentEDDifferentialPrivacy.class);
+            k = (double)dpCriterion.getK();
```
prasser: The minimal class size is extracted at multiple locations, so this comment should be considered wherever applicable:

Generic access to minimal class-size

        // For each
        for (PrivacyCriterion c : this.getPrivacyModels()) {
            if (c.isMinimalClassSizeAvailable()) {
                result = Math.max(result, c.getMinimalClassSize());
            }
        }

In this case: dpCriterion.getMinimalClassSize();

prasser: In the same context: please also overwrite getRiskThresholdJournalist(), getRiskThresholdMarkter() and getRiskThresholdProsecutor() in AbstractDPCriterion and return 1/k in each case.

prasser: Unrelated issue: can we rename the following classes please (if this does not introduce serializability issues):
AbstractEDDifferentialPrivacy -> EDDifferentialPrivacyAbstract
DataDependentEDDifferentialPrivacy -> EDDifferentialPrivacyOptimal

RaffaelBild:
Thanks you for the feedback! I have addressed your comments.

PR #118, opened 2017-07-18
Consider maximal number of records in classification (Sampling size)

eicherj:
If maximal number of records is defined between 0 and the number of rows contained in the dataset, set as sampling size (number of rows to consider for the classification).


eicherj:
Please review.

PR #125, opened 2017-08-08
Add junit test for logistic regression.

eicherj:


src/test/org/deidentifier/arx/test/TestClassification.java
```diff
@@ -0,0 +1,162 @@
+package org.deidentifier.arx.test;
```
prasser: The license header is missing.

PR #129, opened 2017-10-04
Implement classification metric and add to GUI

eicherj:
- Implement classification metric and add to GUI.
- Add functionality to select response variables via GUI.


src/gui/org/deidentifier/arx/gui/resources/Resources.java
```diff
@@ -45,8 +45,11 @@
 
 /**
  * This class provides access to basic resources.
+ *  
+ * Images by FAMFAMFAM, http://www.famfamfam.com/.
```
prasser: Why? None of the icons that existed prior to this PR are from FAMFAMFAM (whatever this is) and I don't think that you needed to use this website to draw a black circle around them, right?

src/gui/org/deidentifier/arx/gui/view/impl/common/ViewDataInput.java
```diff
@@ -199,13 +200,28 @@ public void update(final ModelEvent event) {
...
                     final int index = handle.getColumnIndexOf(attr);
-                    updateHeaderImage(index, definition.getAttributeType(attr));
+                    updateHeaderImage(index, attr, definition);
     
                     // Redraw
                     table.setEnabled(true);
                     table.redraw();
                 }
             }
+        } else if (event.part == ModelPart.SELECTED_ATTRIBUTE) {
```
prasser: I don't understand this block. Why do you need to update the header images after the user has selected a different attribute? Maybe you want to react on updated response variables? If so, you need a specific ModelEvent for this.

prasser: And: you could implement this like it is implement for a change in AttributeType. Store the affected attribute in the even and only update the image for this specific attribute. Then, as far as I can tell, the method ViewData.updateHeaderImages is also not needed!

src/gui/org/deidentifier/arx/gui/view/impl/define/ViewAttributeList.java
```diff
@@ -225,28 +238,24 @@ private void create(final Composite parent) {
...
-                if (model == null || model.getInputConfig() == null || model.getInputConfig().getInput() == null) {
-                    return;
-                }
-                int index = table.getSelectionIndex();
-                if (index >= 0 && index <= model.getInputConfig().getInput().getHandle().getNumColumns()) {
-                    String attribute = model.getInputConfig().getInput().getHandle().getAttributeName(index);
-                    model.setSelectedAttribute(attribute);
-                    controller.update(new ModelEvent(this, ModelPart.SELECTED_ATTRIBUTE, attribute));
-                }
+                updateSelection();
```
prasser: I don't think that this needs to be factored out into a new method!

src/gui/org/deidentifier/arx/gui/view/impl/define/ViewAttributeList.java
```diff
@@ -269,6 +278,31 @@ private void create(final Composite parent) {
...
+                                String responseVariable = model.getInputConfig().getInput().getHandle().getAttributeName(index);
+                                // If already contained, remove
+                                boolean isResponseVariable = true;
+                                if(model.getInputDefinition().isResponseVariable(responseVariable)) {
+                                    isResponseVariable = false;
+                                }
+                                model.getInputDefinition().setResponseVariable(responseVariable, isResponseVariable);
+                                updateEntries();
+                                table.setSelection(index);
+                                updateSelection();
```
prasser: You need to implement a special ModelEvent for this, then you don't need to call table.setSelection, updateSelection etc., right?

src/gui/org/deidentifier/arx/gui/view/impl/define/ViewAttributeList.java
```diff
@@ -269,6 +278,31 @@ private void create(final Composite parent) {
...
+                        if (rect.contains(pt)) {
+                            if (i == 4) {
+                                String responseVariable = model.getInputConfig().getInput().getHandle().getAttributeName(index);
+                                // If already contained, remove
+                                boolean isResponseVariable = true;
+                                if(model.getInputDefinition().isResponseVariable(responseVariable)) {
+                                    isResponseVariable = false;
+                                }
+                                model.getInputDefinition().setResponseVariable(responseVariable, isResponseVariable);
+                                updateEntries();
```
prasser: It would be much better, to directly update the item here instead of calling "updateEntries". Why not just item.setImage(4, IMAGE)?

src/gui/org/deidentifier/arx/gui/view/impl/define/ViewAttributeList.java
```diff
@@ -269,6 +278,31 @@ private void create(final Composite parent) {
...
+                int index = table.getTopIndex();
+                while (index < table.getItemCount()) {
+                    TableItem item = table.getItem(index);
+                    for (int i = 0; i < 5; i++) {
+                        Rectangle rect = item.getBounds(i);
+                        if (rect.contains(pt)) {
+                            if (i == 4) {
+                                String responseVariable = model.getInputConfig().getInput().getHandle().getAttributeName(index);
+                                // If already contained, remove
+                                boolean isResponseVariable = true;
```
prasser: isResponseVariable = !model.getInputDefinition().isResponseVariable(responseVariable);

src/gui/org/deidentifier/arx/gui/view/impl/define/ViewAttributeList.java
```diff
@@ -458,9 +492,11 @@ private void updateEntries() {
             TableItem item = new TableItem(table, SWT.NONE);
             item.setText(new String[] { "", attribute, getDataType(attribute), getDataTypeFormat(attribute) }); //$NON-NLS-1$
             AttributeType type = model.getInputDefinition().getAttributeType(attribute);
-            item.setImage(0, controller.getResources().getImage(type));  
-            if (model.getSelectedAttribute() != null && model.getSelectedAttribute().equals(attribute)) {
```
prasser: Why has the part where the selected attribute is also selected in the table been removed? Please implement a specific ModelEvent as described above, then you don't need to do crazy hacks with selected attributes, which means that you can probably keep this. Please change as little as possible.

src/gui/org/deidentifier/arx/gui/view/impl/define/ViewAttributeList.java
```diff
@@ -483,4 +519,18 @@ private void updateSelectedAttribute(String attribute) {
             }   
         }
     }
+    
+    /**
+     * Update selection
+     */
+    private void updateSelection() {
```
prasser: Probably not needed as a separate method....

src/gui/org/deidentifier/arx/gui/resources/Resources.java
eicherj: That is unrelated to the four new icons. I thought all icons are from this source.

src/gui/org/deidentifier/arx/gui/view/impl/common/ViewDataInput.java
eicherj: My intention was to react on updated response variables by updating the icons in all relevant views. I used the wrong event. Is it correct not to introduce a new event but to use the event ATTRIBUTE_TYPE? Different icons are related to different attribute types and thus this event should have the desired effect for the case when response variables change.

eicherj:
Requested changes have been implemented. Please review.

src/gui/org/deidentifier/arx/gui/view/impl/define/ViewAttributeList.java
```diff
@@ -225,16 +238,20 @@ private void create(final Composite parent) {
...
         column2.setText(Resources.getMessage("ViewAttributeList.1")); //$NON-NLS-1$
-        column2.setWidth("25%", "30px"); //$NON-NLS-1$ //$NON-NLS-2$
+        column2.setWidth("20%", "30px"); //$NON-NLS-1$ //$NON-NLS-2$
         DynamicTableColumn column3 = new DynamicTableColumn(table, SWT.NONE);
         column3.setText(Resources.getMessage("ViewAttributeList.2")); //$NON-NLS-1$
-        column3.setWidth("25%", "30px"); //$NON-NLS-1$ //$NON-NLS-2$
+        column3.setWidth("20%", "30px"); //$NON-NLS-1$ //$NON-NLS-2$
+        DynamicTableColumn column4 = new DynamicTableColumn(table, SWT.NONE);
+        column4.setText(Resources.getMessage("ViewAttributeList.3")); //$NON-NLS-1$
+        column4.setWidth("15%", "30px"); //$NON-NLS-1$ //$NON-NLS-2$
```
prasser: The relative widths (%) of the 5 columns don't add up to 100%...

eicherj: Fixed. Please review once more.

PR #130, opened 2017-10-05
Move common classification params to configuration super-class.

eicherj:
Classification parameters that are used by all implemented classifiers (LR, NB, SVM, RF) were moved from each classifier's configuration to the common super class ARXClassificationConfiguration.

This is the first step in preparation for implementing the classifier selection and possibility to adjust classifier parameters in the GUI.


prasser:
Have you checked whether this breaks backwards compatibility to version 3.6.0. of ARX? To do this, you should change the settings for logistic regression in a 3.6.0 project, store the project and then try to open the project using the version of ARX from this branch/revision...

eicherj:
This refactoring seems to be not backwards compatible with version 3.6.0:
I have launched ARX 3.6.0, created a new project, change the relevant properties in the logistic regression settings (maximal number of records, etc.), and then saved the project.
Then I have launched ARX 3.7.0 directly from this branches source code, opened the project and the settings are empty. See attached screenshot. Is that fixable or do I need to revert these changes?


image

PR #136, opened 2017-10-26
Fix mouseevent action in Attribute metadata

eicherj:
- show context menu if right click on columns Data type or Format
- toggle response variable if left click on colum Response variables


eicherj:
Please review again.

PR #137, opened 2017-10-27
 Add functionality to update attribute type of all attributes at once.

eicherj:
In the Data Transformation/Attribute metadata view there are now three additional buttons (for qis, identifying, insensitive) in the title bar which enable the user to update the attribute type of all attributes at once.


src/gui/org/deidentifier/arx/gui/view/impl/define/ViewAttributeTransformation.java
```diff
@@ -348,6 +349,39 @@ public void update(final ModelEvent event) {
             }
         }
     }
+    
+    /**
+     * Update attribute type of all attributes.
+     * @param typeNew
+     */
+    public void actionUpdateAttributeTypes(AttributeType typeNew) {
```
prasser: Please implement all this within the class LayoutAttributeSettings.

eicherj:
Please review.

src/gui/org/deidentifier/arx/gui/view/impl/define/ViewAttributeTransformation.java
```diff
@@ -150,7 +151,7 @@ public ViewAttributeTransformation(final Composite parent,
         cmbType.addSelectionListener(new SelectionAdapter() {
             @Override
             public void widgetSelected(final SelectionEvent arg0) {
-                actionAttributeTypeChanged();
+                actionAttributeTypeChanged(attribute, COMBO1_TYPES[cmbType.getSelectionIndex()]);
```
prasser: Your are now lacking the sanity checks that were implemented previously. E.g. if ((cmbType.getSelectionIndex() != -1) && (attribute != null)) {

PR #139, opened 2017-11-03
Update attribute type of selected attribute in GUI.

eicherj:
When changing all attribute types at once via GUI the currently selected attribute's type has to be changed in the combo box.

### DIFF
--- a/src/gui/org/deidentifier/arx/gui/view/impl/define/ViewAttributeTransformation.java
+++ b/src/gui/org/deidentifier/arx/gui/view/impl/define/ViewAttributeTransformation.java
@@ -545,6 +545,15 @@ public class ViewAttributeTransformation implements IView {
                     AttributeType type = model.getInputDefinition().getAttributeType(attribute);
                     // Type changed
                     if (type != typeNew) {
+                        // Change selection in combo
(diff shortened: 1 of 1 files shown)


src/gui/org/deidentifier/arx/gui/view/impl/define/ViewAttributeTransformation.java
```diff
@@ -545,6 +545,15 @@ public void actionUpdateAttributeTypes(AttributeType typeNew) {
                     AttributeType type = model.getInputDefinition().getAttributeType(attribute);
                     // Type changed
                     if (type != typeNew) {
+                        // Change selection in combo
+                        if (attribute.equals(model.getSelectedAttribute())) {
```
prasser: Wouldn't it be more intuitive to test for equality with "this.attribute"? AFAICT "this.attribute" and "model.getSelectedAttribute()" should always be equal, however, testing against the former seems to be more intuitive as we are manipulating the view represented by the current object...

eicherj: It probably is more intuitive, so I changed it. Please review.

PR #150, opened 2017-12-03
Integrate reliable parameter calculation into the differential privacy model

RaffaelBild:


prasser:
Please reformulate this PR against the new code that simplified the initialization of EDDifferentialPrivacy

PR #158, opened 2018-01-15
Classification changes

eicherj:
Update target variables (image) after response variables have changed.
* Add response variables to the State
* Add listener/update for ModelPart.RESPONSE_VARIABLES
* Choose target image based on type and whether isResponseVariable

Set classification parameter after anonymization process is completed.
* Set all quasi-identifying attributes as feature variables
* Set all response variables, if any, as target variables

And minor changes.


eicherj:
Please review again.

PR #161, opened 2018-02-05
Implemented test for managing and assessing re-identification risks

ElMuto:
- implemented the concatenation of two anonymizations (with and without records at risk) with overlapping quasi-identifiers
- implemented the validation of the assessment results with and without wildcard match


ElMuto:
deprecates

PR #176, opened 2018-03-05
Handle: Saving projects with errors.

eicherj:
Try saving project to temporary file and only if successful save to
actual target file.


eicherj:
Please review again.

PR #177, opened 2018-03-05
Bugfix serialization in ARXFeatureScaling

eicherj:
Remove net.objecthunter.exp4j.Expression from class member since this
class is not serializable.


eicherj:
Please review again.

Issue #185, opened 2018-03-21
Question regarding "HierarchyBuilderRedactionBased.java"


victorxwu:
## I had a problem when creating generalization hierarchy through "using masking".

i.e. Let's say we have a string "helloworld". And let's assume I want to mask it into something like "*****world".

**So this is what I would do to achieve the result using masking hierarchy:**
_1. align items to the right
2. mask characters left to right_

However, the output would become "*****hello". 

In the file "[HierarchyBuilderRedactionBased.java](https://github.com/arx-deidentifier/arx/blob/master/src/main/org/deidentifier/arx/aggregates/HierarchyBuilderRedactionBased.java#L490)", there are codes look like this:
```
// Build result
        this.result = new String[base.length][length + 1];
        for (int i=0; i<base.length; i++){
            result[i] = new String[length + 1];
            result[i][0] = data[i];
            for (int j=1; j<length + 1; j++){
                String redact = redaction.substring(0, j);
                if (redactionOrder == Order.RIGHT_TO_LEFT) {
                    result[i][j] =  base[i].substring(0, length - j) + redact;
                } else {
                    result[i][j] =  redact + base[i].substring(0, length - j);
                }
            }
        }
```

May I ask if ARX is doing it intentionally? Since I assume masking from left to right means the output should be "*****world" instead of "*****hello", I am a bit confused about the last line of your code. 

I think to achieve "*****world" masking, the last line of the code block would be something like:
```result[i][j] =  redact + base[i].substring(j, length);```

Did I misunderstand the masking process? Is it possible to mask "helloworld" into "*****world" through ARX? May I please get some insight on that?

Thanks,

Wu

prasser:
Dear Wu,

thanks for your interest in ARX.

You are right. This is a bug (we typically use right to left, that's why we never noticed the problem). Can you provide a pull-request?

Thanks for reporting this! 
Fabian

victorxwu:
Dear Fabian,

Thank you very much for you quick reply.
This is my first time using pull-request on GitHub. I hope it is working as intended.


Thank you,

Wu

prasser:
Fix merged.

PR #186, opened 2018-03-21
Update HierarchyBuilderRedactionBased.java

victorxwu:
Attempt to fix the issue with malfunctioning "masking from left to right" for creating generalization hierarchy on ARX.


prasser:
This fixes #185. Looks good. Thanks!

PR #195, opened 2018-05-17
Use maven build and support intellij

nartz:
This is an attempt to 'mavenize' the ARX project. Still a WIP, waiting for feedback. This should probably live on a separate feature branch until master is ready for a change like this. Still working on updating the tests, although most of them are passing.

The main changes involved are:

1. Move the project to conform to the standard/default maven directory structure; this is sort of a large change, and any existing non-merged branches may have a large diff
[see here](https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html).
- Move the 'gui' and 'examples' directories into the main folder as subfolders (src/main/java/...)
- Move the tests into src/test/java/...
- Move resources into their respective 'resources' folders

2. Create a pom.xml 
- dependencies found on maven central were added to the pom and linked
- dependencies not found on maven central were left in the lib folder, using a 'system' scope (this probably should be undone over time as dependencies are updated, eventually all should be managed by maven central or a separate nexus)
- supports generating jar files with correct dependencies with the same output names (`mvn install`, or to build for a different platform, `mvn install -P<platformName>` i.e. `mvn install -Pwin`)

3. Add some intellij files that will be useful for project setup, but while using eclipse can be ignored.

Tested only on OSX thus far.


prasser:
Perfect!

prasser:
Dear Nate,

as you might have seen, we released ARX 3.7.0 a few weeks ago. So, now is the right time to merge (parts of?) the changes proposed in this PR. What do you think. Can you:

(1) Create a PR that just adds support for IntelliJ?

and/or

(2) Create a PR that mavenizes the project, but is minimally invasive?

If the answer to (1) and (2) is "no", we need to find another way. I would really like to merge this...

Thanks
Fabian

prasser:
Subsumed by  #206

PR #202, opened 2018-07-11
Add Intellij Configuration

marklackey:
Add run configurations and required config files in the .idea folder


.idea/runConfigurations/arx_3_7_0_osx_64_main.xml
```diff
@@ -0,0 +1,10 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="arx-3.7.0-osx-64-main" type="JarApplication" factoryName="JAR Application">
```
prasser: Does it really make sense to include this configuration, which points to a specific build of ARX? I suggest to remove this file. If this is not possible, please explain why.

.idea/copyright/rls.xml
```diff
@@ -0,0 +1,6 @@
+<component name="CopyrightManager">
+  <copyright>
+    <option name="notice" value="Copyright (c) &amp;#36;today.year. REAL LIFE SCIENCES INC" />
```
prasser: Does it really make sense to include a RLS copyright notice? I am not an Intellij user, so I don't know what it is used for, but I have the feeling that it could (should?) be removed.

prasser: Does it really make sense to include a RLS copyright statement here? I am not an Intellij user, so I don't know what this is used for, but I have the feeling that this could (should?) be removed?

.idea/runConfigurations/arx_3_7_0_osx_64_main.xml
prasser: This configuration points to a specific build of ARX. I thinkt that it should be removed. If not, please explain why it is needed.

nartz: This run configuration is if a user wanted to run a jar in intellij, after having run the ant build - this makes it easier. Of course they can use the other builds to test, but that would be testing using intellij's automake. However, as you point out, because we put the version names directly in the jar, it makes it hard here to specify the jar name. Removing is probably the right option.

.idea/copyright/rls.xml
nartz: agree we can remove

.idea/runConfigurations/arx_3_7_0_osx_64_main.xml
marklackey: I agree we can remove this. It does provide some useful configuration details for the developer. Is there a "developer readme" or other developer documentation somewhere where we could add some information about running and/or debugging the jar? E.g. to run and allow for debugging, these parameters are required (but perhaps not obvious): `-Xdebug -XstartOnFirstThread -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005`

prasser: I see. I think that there are two options:

(1) Create three such configurations, one for each build of ARX (Windows, Linux/GTK and OSX). The configurations should be independent of a release number (e.g. called "arx-current-release-osx"). If possible, I suggest to extract the current release number from ARX's build.xml (property "arx.version"). If this is not possible, it could be possible to include wildcards or the like into the reference to the jar-files (maybe "$PROJECT_DIR$/jars/arx-*-osx-64.jar" would work?)

(2) Create a developer readme. I would be happy to include such a file anyway. This could be implemented instead of (1) or in addition.

.idea/copyright/rls.xml
prasser: I don't know how this is used by IntelliJ, so this question is a bit hard to answer. If it is unusual to not include such a copyright statement, I would indeed suggest to replace the value with ARX's standard copyright notice ("Copyright (c) &amp;#36;today.year. Fabian Prasser and contributors"). If this is implemented, please also change the name of the file, e.g. to arx.xml, and insert a reasonable value into "<option name="myName" value="rls" /". Can we also include a reference to the Apache 2 license here?

.idea/runConfigurations/arx_3_7_0_osx_64_main.xml
nartz: I think we should just drop this configuration for now - we could do this - but running a jar is extremely easy with intellij, with a few clicks one can generate this. The more important ones we've already included which allow using native intellij's automake.

prasser:
I see. Sounds good to me. Thanks, Fabian

Am 16. Juli 2018 17:44:04 MESZ schrieb nartz <notifications@github.com>:
>nartz commented on this pull request.
>
>
>
>> @@ -0,0 +1,10 @@
>+<component name="ProjectRunConfigurationManager">
>+  <configuration default="false" name="arx-3.7.0-osx-64-main"
>type="JarApplication" factoryName="JAR Application">
>
>I think we should just drop this configuration for now - we could do
>this - but running a jar is extremely easy with intellij, with a few
>clicks one can generate this. The more important ones we've already
>included which allow using native intellij's automake.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/arx-deidentifier/arx/pull/202#discussion_r202731775

-- 
Dr. Fabian Prasser

Institute of Medical Informatics, Statistics and Epidemiology
University Hospital rechts der Isar
Technical University of Munich

Grillparzerstr. 18 (3rd floor)
81675 Munich
Germany

Tel: +49 89 4140 4328
Fax: +49 89 4140 4850
Web: www.imse.med.tum.de

.idea/runConfigurations/arx_3_7_0_osx_64_main.xml
prasser: I see. Sounds good to me. Thanks, Fabian

PR #206, opened 2018-07-24
Tek 1064 create maven build

marklackey:
This commit includes three major pieces:

1) A pom file that describes the build (pom.xml)
2) A local maven repository (local-maven-repo/*)
3) Intellij Maven Integration config files (in the .idea folder)

There are also two points of interest @prasser :

1) We are compiling to/with Java 8 because our environment use Java 8, and because of course Java 8 has some nice features. It seemed like the previous build was 1.7, but that 1.8 was okay as well. Please let me know if we need to target 1.7.
2) This section in the pom is currently commented out, without impacting the build so far as I can tell, and I can't find it being used in the project. So, a) is it possible this is no longer needed? or b) could we perhaps upgrade to a more recent version that is still supported?
`        
        
        
        
        
        ` 

- We have tested that osx-64, gtk-64, and win32-64 builds run, and we have run the test suite on osx. We suggest that 32-bit osx builds be retired--we were unable to test this version. (The last 32-bit mac hardware was manufactured in 2006.) We have not tested 32-bit windows.
- We attempted to find all dependencies in public maven repositories, but some libraries were not available, hence the local maven repository.
- All the typical maven lifecycle phases work as expected
- This maven build includes Profiles. If run on a given platform, maven should detect the platform and "activate" that Profile. The profile can also be specified on the command line. (I had to specify the profile on Linux).


prasser:
Thanks, Mark! Java 8 is fine and I'll take a look at the other open questions later.

There are two aspects of this, that I don't like at the moment:

(1) Do we really need a local maven repo? I would strongly prefer to add as little to the git repo as necessary. Can we not just reference the existing files in "lib" where needed? Something along the following lines should work, or not?

<dependency>
	<groupId>X</groupId>
	<artifactId>Y</artifactId>
	<version>Z</version>
	<scope>system</scope>
	<systemPath>${project.basedir}/lib/X/Y.jar</systemPath>
</dependency>

(2) The ".idea" folder contains references to many libraries that are not needed by ARX (e.g. Apache Solr). These should be removed.

marklackey:
Hi Fabian.
1) We don't like having to add this. And in general, would prefer to avoid having libraries in the repo at all. We can't simply point maven to the lib dir. There are configs and checksums that maven needs. However, along these lines, it should work to do this in reverse: point ant to the maven repo. It would simply require altering the paths in the build.xml. 
The other option is to use a hosted maven repo. This one is free for OSS: https://central.sonatype.org/pages/ossrh-guide.html It requires signing up. There are other options, such as running your own nexus instance, or other ideas like using dropbox: https://jitblog.net/use-dropbox-to-host-a-maven-repository/

2) Apache Solr is a dependency of the apache mahout library in use in the project. It seems like Intellij needs to have dependencies of dependencies mapped out in its config.

marklackey:
Also, @prasser , I'm noticing these errors when running on Linux. ARX loads and seems to run, but generates this error multiple times.
On Ubuntu 14.04:
`(ARX Anonymization Tool:7870): GLib-GObject-WARNING **: /build/buildd/glib2.0-2.40.2/./gobject/gclosure.c:688: unable to remove uninstalled invalidation notifier: 0x7f2e0562be00 (0x7f2e34dcfc40)`
Ubuntu 17.10:
`(ARX Anonymization Tool:6055): GLib-GObject-CRITICAL **: g_closure_add_invalidate_notifier: assertion 'closure->n_inotifiers < CLOSURE_MAX_N_INOTIFIERS' failed`

`(ARX Anonymization Tool:6055): GLib-GObject-WARNING **: ../../../../gobject/gclosure.c:724: unable to remove uninstalled invalidation notifier: 0x7fb2677e4580 (0x7fb298589810)`

prasser:
Hi Mark,

Re. 1: I'm still not convinced. We use the configuration that I proposed above (only jars in "lib" folder, dependencies with scope "system" in the pom) in some of our internal projects and it works just fine. So I don't think that Maven always needs those configs and checksums. @ElMuto any thoughts on this?

Also, to be completely clear: in the mid-term we also want to avoid having any libraries in the repository at all and we will work on this by making sure that all dependencies are available via public maven repos. We just don't have a lot of resources for working on this at the moment and that's why I was hoping for a smooth transition process.

Re. 2: It would be great if we could get rid of these transitive dependencies. IntelliJ seems to provide options to prevent transitive dependencies from being laid out in the config (see https://www.jetbrains.com/help/idea/maven-support.html#view_transitive_dependency_maven).

In summary, my major concern with this PR is that it involves changes to more than 350 files, most of which are additions. This should be avoided at all cost, as it can easily turn into a maintenance nightmare.

Re. warnings and errors on Linux/GTK: This is a known issue which has existed for years. I don't know of any fix at the moment. Even updating to the latest versions of SWT and JFace didn't make the messages go away when I last tried a few weeks ago. They don't seem to have any effect on ARX's functionality, however.

Thanks!
Fabian

marklackey:
Fabian,

Got it on the Linux errors. Thanks.

I'll respond again later regarding the intellij files, but one point on using <scope>system</scope>: It doesn't seem to work with the maven shade plugin, which we use to build the uber jar. That's why I moved from system scope to a more conventional repository approach.

-Mark

nartz:
I think its important to come from the idea that not many projects these days support BOTH ant and maven, so some of the things we are doing we might have to compromise on.

1. @prasser You are right - we can use 'system' level jars, which if you saw I was doing in my original branch. After more investigation there are a number of negatives to doing it this way, especially in a maven project. For instance, please see: [https://stackoverflow.com/questions/3642023/having-a-3rd-party-jar-included-in-maven-shaded-jar-without-adding-it-to-local-r](https://stackoverflow.com/questions/3642023/having-a-3rd-party-jar-included-in-maven-shaded-jar-without-adding-it-to-local-r)

The *maven*ey way to do this is as we have, by creating a local repo, which works much better as a mechanism with all of the dependency related maven plugins and IDEs. One example, the plugin we use to build one 'master' jar is the maven shade plugin, which does not work with system level jars (ignores them - see link).

This being the main 'maven' plugin to do this, the maven community sort of prefers to use the shade plugin. That being said, there are other ways, and some 3rd party jars to build a master jar - these essentially look at all the jars in the system scope and 'copy' them on the fly at compile time into a maven repository for it to work.

2. I think @marklackey is working on removing non-needed dependencies and pruning them down - however, often as shown, there are a lot of other dependencies included in the various libraries from maven central. @prasser we have done our best as a start to link to maven central - but we might have selected too broadly or gotten it wrong - the more details suggestion you can give the better.

The great thing about this is that now we have a pom, these are much quicker changes to changing / upgrading dependencies.

I'm not sure we should get *too* hung up on including a few extra dependencies in the pom, they can easily be removed as we go along. It might even be better to have a separate story after this is merged to do some housecleaning which I imagine might need to be done anyway (seems like there might have been a few jars in lib getting very little use anyway inside ARX).

prasser:
Re 1. Thanks for the link, I'm now convinced that we should keep the local repo and point into it from our ant build file. A few follow-up questions: 

(a) Have you used a tool to create it? I'm asking, because I am trying to understand how we could clean it up a little bit. For example, I suggest to change a few of the group-ids of the jars in the local repo, which in turn seems to require changes to the directory structure etc. Looks pretty complicated without tool support (maybe it's easier if you perform these changes before we merge this, I will add comments).

(b) Can we rename "local-maven-repo" to "lib" (basically merging both folders), or is this name also a convention or will maven not be happy with the other files and folders hanging around in "lib" for the moment?

(c) The local maven repository contains many dependencies that should also be available via public repositories. Is this on purpose (e.g. to allow us to reference them from the ant build file) or is there another reason for this?

Re 2. I'm not sure what you mean by "the more details suggestion you can give the better". If you want suggestions on which libraries need to be included I would refer you to the "lib" folder. If you want suggestions on how this can be narrowed down: I'll try, but I'm (obviously) not too familiar with the matter. Also: you are right - we might want to perform housekeeping separately.

One last question: The file "arx.iml" has been moved from ".idea" to project root. I would really like to keep it in ".idea", if possible (although I know that project root seems to be the standard location). Is this doable?

Thanks a lot for your efforts and I'm sorry for being a PITA!
Fabian

pom.xml
```diff
@@ -0,0 +1,459 @@
...
+            <artifactId>common</artifactId>
+            <version>3.6.100</version>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse</groupId>
+            <artifactId>jface</artifactId>
+            <version>3.8.102</version>
+        </dependency>
+        <dependency>
+            <groupId>jhc</groupId>
```
prasser: groupId = de.linearbits

pom.xml
```diff
@@ -0,0 +1,459 @@
...
+            <artifactId>libjhc</artifactId>
+            <version>swt</version>
+        </dependency>
+        <dependency>
+            <groupId>jhpl</groupId>
```
prasser: groupId = de.linearbits

pom.xml
```diff
@@ -0,0 +1,459 @@
...
+            <artifactId>jtds</artifactId>
+            <version>1.3.1</version>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.nebula.widgets.nattable</groupId>
+            <artifactId>core</artifactId>
+            <version>1.1.0</version>
+        </dependency>
+        <dependency>
+            <groupId>newtonraphson</groupId>
```
prasser: groupId = de.linearbits

pom.xml
```diff
@@ -0,0 +1,459 @@
...
+            <artifactId>newtonraphson</artifactId>
+            <version>0.0.1</version>
+        </dependency>
+        <dependency>
+            <groupId>objectselector</groupId>
```
prasser: groupId = de.linearbits

pom.xml
```diff
@@ -0,0 +1,459 @@
...
+            <artifactId>ext</artifactId>
+            <version>0.8.0.v20120301</version>
+        </dependency>
+        <dependency>
+            <groupId>org.swtchart</groupId>
+            <artifactId>swtchart</artifactId>
+            <version>0.8.0.v20120301</version>
+        </dependency>
+        <dependency>
+            <groupId>swtchoicesdialog</groupId>
```
prasser: groupId = de.linearbits.swt

pom.xml
```diff
@@ -0,0 +1,459 @@
...
+            <artifactId>swtchoicesdialog</artifactId>
+            <version>0.0.1</version>
+        </dependency>
+        <dependency>
+            <groupId>swtknob</groupId>
```
prasser: groupId = de.linearbits.swt

pom.xml
```diff
@@ -0,0 +1,459 @@
...
+            <artifactId>swtknob</artifactId>
+            <version>1.0.0</version>
+        </dependency>
+        <dependency>
+            <groupId>swtpreferences</groupId>
```
prasser: groupId = de.linearbits.swt

pom.xml
```diff
@@ -0,0 +1,459 @@
...
+            <artifactId>swtpreferences</artifactId>
+            <version>0.0.1</version>
+        </dependency>
+        <dependency>
+            <groupId>swtrangeslider</groupId>
```
prasser: groupId = de.linearbits.swt

pom.xml
```diff
@@ -0,0 +1,459 @@
...
+            <artifactId>swtrangeslider</artifactId>
+            <version>0.0.1</version>
+        </dependency>
+        <dependency>
+            <groupId>swtsimplebrowser</groupId>
```
prasser: groupId = de.linearbits.swt

pom.xml
```diff
@@ -0,0 +1,459 @@
...
+            <artifactId>swtsimplebrowser</artifactId>
+            <version>0.0.1</version>
+        </dependency>
+        <dependency>
+            <groupId>swttable</groupId>
```
prasser: groupId = de.linearbits.swt

pom.xml
```diff
@@ -0,0 +1,459 @@
...
+            <artifactId>swttable</artifactId>
+            <version>0.0.1</version>
+        </dependency>
+        <dependency>
+            <groupId>swttiles</groupId>
```
prasser: groupId = de.linearbits.swt

marklackey:
- removed transitive dependencies
- renamed group id's as requested
- moved maven repo to lib folder
- created 'ant' folder inside 'lib' folder to avoid confusion
- modified build.xml to point to ant folder
- we have created a few scripts to generate maven deploy commands to install the jar's into the local repository
- we only included libraries in the local repo that we couldn't find in public repo's.
- it may be possible to move the arx.iml file to the .idea directory. however, it would require some work... when i move it there, the project no longer loads properly in Intellij

prasser:
Thanks for even fixing the ant-file, which I did not expect! The following issue needs to be fixed, however:
- The ojdbc driver has been unpacked (1500+ .class files in lib/ant/ojdbc).

 I have one related question:
- The local maven repository contains many dependencies that should also be available via public repositories. Is this on purpose (e.g. to allow us to reference them from the ant build file) or is there another reason for this?

marklackey:
When Nate created his initial maven build, he looked for each library in public maven repo's. The one's he couldn't find, we included in the local repo.

prasser:
Thanks for the feedback. I (think I) found the first library that I tried: 

https://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22org.apache.poi%22%20AND%20a%3A%22poi-ooxml%22

Wouldn't this work?

marklackey:
We're not including that library in the local maven repo. (I.e. it's not in lib/org/apache...). It's simply listed in the pom file as a dependency and is found with jitpack.io.

prasser:
Ah, sorry, my mistake. I misinterpreted the files contained in ".idea". I'm having a hard time reviewing this PR, as it is so big. I'm currently trying the Octotree plugin for Chrome, but it is not listing the changes in "lib" for some reason....

prasser:
Ok, I've now taken some time to review this as best as I can at the moment: I thank you a lot for your efforts and will definitely pull this! 

It would be great if we could fix two minor issues before that:

1) The dependency "swtknob" should also have the group-id "de.linearbits.swt". I think I forgot to suggest this change and I'm sorry for that.

2) Regarding Mark's comments on ojdbc7.jar above: I think that the pom already contains a reference to the official dependency (cf. group-id: com.oracle) and you have already "upgrade[d] to a more recent version that is still supported" as suggested. This is the Oracle JDBC driver, which is loaded dynamically when users try to access an Oracle database from the GUI. Consequently, removing it doesn't break the build but it should definitely be compiled into the final jars. AFAIK, we can simply remove the section that is commented out in the pom, right?

Thanks
Fabian

PR #213, opened 2018-08-08
 Improve handling of initialization for the differential privacy model 

RaffaelBild:


src/main/org/deidentifier/arx/criteria/EDDifferentialPrivacy.java
```diff
@@ -58,11 +58,11 @@
     /** Parameter */
     private DataSubset               subset;
     /** Parameter */
-    private transient DataManager    manager;
-    /** Parameter */
     private transient boolean        deterministic    = false;
     /** Parameter */
     private DataGeneralizationScheme generalization;
+    /** Indicates if this instance was already initialized */
```
prasser: Comment on why this variable is transient.

src/main/org/deidentifier/arx/criteria/EDDifferentialPrivacy.java
```diff
@@ -160,62 +160,59 @@ public int getRequirements(){
...
+    @Override
     public void initialize(DataManager manager, ARXConfiguration config){
         
-        // Needed for consistent de-serialization. We need to call this
-        // method in the constructor of the class DataManager. The following
-        // condition should hold, when this constructor is called during 
-        // de-serialization, when we must not change the subset.
-        if (subset != null && this.manager == null) {
-            this.manager = manager;
+        if (config != null) {
```
prasser: Because  this is the only privacy model that dynamically constructs a subset, we need to call this method twice. This is the second call and can be ignored.

src/main/org/deidentifier/arx/criteria/EDDifferentialPrivacy.java
prasser: An instance is initialized when it is used to anonymize a dataset. When this model is used more than once, a new subset needs to be drawn before each use (i.e. when "initialized=true". The field is transient, because we need to preserve the subset, when the model is loaded from a project file (indicated by "initialized=false").

src/main/org/deidentifier/arx/criteria/EDDifferentialPrivacy.java
```diff
@@ -160,62 +160,59 @@ public int getRequirements(){
...
-        if (deterministic) {
-            random = new Random(0xDEADBEEF);
-        } else {
-            random = new SecureRandom();
-        }
+        // Perform random sampling iff the subset is null
+        // or if this method has already been called.
+        // Otherwise, we know that this instance was just de-serialized
+        // and that no subsequent call of anonymize() has occured.
+        if (subset == null || initialized) {
```
prasser: Perform random sampling iff the model is used for the first time (subset == null) or when it used again (initialized == true). We don't perform random sampling when the model has been deserialized (subset will be !=null and initialized will be false).

src/main/org/deidentifier/arx/criteria/EDDifferentialPrivacy.java
```diff
@@ -160,62 +160,59 @@ public int getRequirements(){
...
+
+            // Create RNG
+            Random random;
```
prasser: Ternary operator

src/main/org/deidentifier/arx/criteria/EDDifferentialPrivacy.java
```diff
@@ -238,8 +235,8 @@ public ElementData render() {
         ElementData result = new ElementData("Differential privacy");
         result.addProperty("Epsilon", epsilon);
         result.addProperty("Delta", delta);
-        result.addProperty("Uniqueness threshold (k)", k);
-        result.addProperty("Sampling probability (beta)", beta);
+        result.addProperty("Uniqueness threshold (k)", getK());
```
prasser: Catch exception

RaffaelBild:
Thanks for the feedback, I have performed the requested changes.

PR #215, opened 2018-08-28
Prepare integration of data-dependent differential privacy

RaffaelBild:


src/main/org/deidentifier/arx/ARXAnonymizer.java
```diff
@@ -544,6 +545,27 @@ private void checkBeforeEncoding(final DataHandle handle, final ARXConfiguration
             if (!definition.getQuasiIdentifiersWithMicroaggregation().isEmpty()) {
                 throw new IllegalArgumentException("Differential privacy must not be combined with micro-aggregation");
             }
+            EDDifferentialPrivacy edpModel = config.getPrivacyModel(EDDifferentialPrivacy.class);
+            if (edpModel.getEpsilon() <= 0d) {
```
prasser: It would actually make sense to check for correctness in parameterizations in the constructor of the privacy model. However, this is not implemented consistently in ARX, unfortunately. Some models perform checks within their constructors, but many checks are also performed in this method. Consequently, this can remain as-is for now, but please file in issue so that we can fix this for all models in the future.

src/main/org/deidentifier/arx/ARXAnonymizer.java
```diff
@@ -544,6 +545,27 @@ private void checkBeforeEncoding(final DataHandle handle, final ARXConfiguration
...
+                throw new IllegalArgumentException("The privacy budget must be > 0");
+            }
+            if (edpModel.getDelta() <= 0d) {
```
prasser: Shouldn't we also check for Delta>1 ?

src/main/org/deidentifier/arx/ARXConfiguration.java
```diff
@@ -552,6 +560,30 @@ public ARXCostBenefitConfiguration getCostBenefitConfiguration() {
...
+    }
+    
+    /**
+     * Returns the number of steps to use for the data-dependent
+     * differential privacy search algorithm. The default is 100.
+     * @return
+     */
+    public int getDPSearchStepNumber() {
+        if (this.dpSearchStepNumber == null) {
+            this.dpSearchStepNumber = 100;
```
prasser: Wasn't 300 recommended as the default parameterization in the SafePub paper?

src/main/org/deidentifier/arx/ARXLattice.java
```diff
@@ -1196,6 +1202,36 @@ private void build(final Transformation optimum, Map<String, Integer> headermap)
...
         createRelationships(solutions, solutions.getBottom().getIdentifier());
     }
+    
+    /**
+     * Build an ARX lattice for a solution space with only one element
+     * @param solutions 
+     * @param optimum
+     * @param headermap
+     */
+    private void buildSingle(SolutionSpace solutions, final Transformation optimum, Map<String, Integer> headermap) {
```
prasser: Arent't we allowed to select any solution that we checked after executing the differentially private search process? SafePub deterministically selects the optimum. It might be OK, to allow users to manually select any pivot element, meaning that it is not necessary to restrict the lattice to only one element. However, I am unsure about this...

src/main/org/deidentifier/arx/algorithm/DataDependentEDDPAlgorithm.java
```diff
@@ -0,0 +1,183 @@
...
+import org.deidentifier.arx.framework.lattice.Transformation;
+import org.deidentifier.arx.metric.v2.ILScore;
+import org.deidentifier.arx.reliability.IntervalArithmeticDouble;
+import org.deidentifier.arx.reliability.IntervalArithmeticException;
+
+import cern.colt.list.LongArrayList;
+import de.linearbits.jhpl.PredictiveProperty;
+
+/**
+ * This class implements the search algorithm used with data-dependent differential privacy
```
prasser: Please reference the SafePub paper

src/main/org/deidentifier/arx/algorithm/DataDependentEDDPAlgorithm.java
```diff
@@ -0,0 +1,183 @@
...
+        this.checker.getHistory().setStorageStrategy(StorageStrategy.ALL);
+        this.propertyChecked = space.getPropertyChecked();
+        this.solutionSpace.setAnonymityPropertyPredictable(false);
+        this.steps = steps;
+
+        double epsilonPerStep;
+        IntervalArithmeticDouble arithmetic = new IntervalArithmeticDouble();
+        try {
+            epsilonPerStep = arithmetic.div(arithmetic.createInterval(epsilonSearch), arithmetic.createInterval(steps)).lower;
+            exponentialMechanism = new ExponentialMechanismReliable<Long>(epsilonPerStep, deterministic);
```
prasser: We don't include a non-reliable variant of the Expontential Mechanism, right? This means that the class ExponentialMechanismReliable can be renamed to ExponentialMechanism.

src/main/org/deidentifier/arx/algorithm/DataDependentEDDPAlgorithm.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+    }
+
+    /**
+     * Executes the exponential mechanism
+     * @param transformationIDToScore
+     * @return
+     */
+    private long executeExponentialMechanism(Map<Long, ILScore> transformationIDToScore) {
+
```
prasser: Please document what this method is doing by adding comments

src/main/org/deidentifier/arx/criteria/EDDifferentialPrivacy.java
```diff
@@ -69,8 +69,8 @@
      */
```
prasser: Please add a reference to the SafePub paper to the header of this class.

src/main/org/deidentifier/arx/criteria/EDDifferentialPrivacy.java
```diff
@@ -169,18 +178,12 @@ public int getRequirements(){
...
-        if (config != null) {
-            // Because this is the only privacy model that dynamically
-            // constructs a subset, we need to call this method twice.
-            // This is the second call and can be ignored.
-            return;
-        }
-        
         // Set beta and k if required
         if (beta < 0) {
+            double epsilonAnon = epsilon - (isDataDependent() ? config.getDPSearchBudget() : 0d);
```
prasser: Seems to not be a reliable calculation, although errors are obviously very small.

src/main/org/deidentifier/arx/dp/ExponentialMechanismReliable.java
```diff
@@ -0,0 +1,245 @@
...
+
+    /** A cache mapping a pair of exponents (e_1,e_2) to n^{e_1} / d^{e_2} used to increase performance */
+    private Map<Pair<Integer, Integer>, BigInteger> productCache;
+
+    /**
+     * Creates a new instance
+     * @param epsilon
+     * @throws IntervalArithmeticException 
+     */
+    public ExponentialMechanismReliable(double epsilon) throws IntervalArithmeticException {
```
prasser: Please rename to ExponentialMechanism

src/main/org/deidentifier/arx/dp/ExponentialMechanismReliable.java
```diff
@@ -0,0 +1,245 @@
...
+        // Assure that score is within the range of numbers which can be processed
+        if (fraction.compareTo(new BigFraction(Integer.MAX_VALUE)) == 1 || (fraction.subtract(1)).compareTo(new BigFraction(Integer.MIN_VALUE)) == -1) {
+            throw new IllegalArgumentException("The absolute value of " + fraction + " is too big to be processed");
+        }
+
+        // Extract the whole number part of the fraction 
+        int result = fraction.intValue();
+
+        // Correct direction of rounding if necessary
+        if (result < 0 && result != fraction.intValue()) {
```
prasser: Please explain this in more detail. It seems weird.

src/main/org/deidentifier/arx/dp/ExponentialMechanismReliable.java
```diff
@@ -0,0 +1,245 @@
...
+     * Returns a random BigInteger in the interval [0, limit).
+     * The parameter limit has to be >= 1, otherwise an illegal argument exception is thrown.
+     * @param limit
+     * @return
+     */
+    private BigInteger getRandomBigInteger(BigInteger limit) {
+
+        // Check the argument
+        if (limit.compareTo(BigInteger.ONE) == -1) {
+            throw new IllegalArgumentException("limit has to be greater than or equal to one");
```
prasser: Please start messages with upercase letters.

src/main/org/deidentifier/arx/metric/v2/DomainShareReliable.java
```diff
@@ -0,0 +1,185 @@
...
+                int value = strategy[level];
+                String keyString = encodedValues[value];
+                BigFraction share = new BigFraction(map.get(keyString), rawHierarchy.length);
+                BigFraction stored = shares[value];
+
+                // If duplicate
+                if (stored != NOT_AVAILABLE) {
+
+                    // If same share, simply continue
+                    if (stored == share) {
```
prasser: Please double check. I am not sure(!), but stored.equals(share) may be what you are looking for. This compares references.

src/main/org/deidentifier/arx/metric/v2/ILScore.java
```diff
@@ -0,0 +1,130 @@
...
+ * limitations under the License.
+ */
+
+package org.deidentifier.arx.metric.v2;
+
+import org.apache.commons.math3.fraction.BigFraction;
+import org.deidentifier.arx.metric.InformationLoss;
+
+/**
+ * This class implements a score value in the form of an information loss,
```
prasser: "score value in the form of an information loss" is weird.

src/main/org/deidentifier/arx/metric/v2/ILScore.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+    
+    @Override
+    public InformationLoss<BigFraction> clone() {
+        return new ILScore(getValue());
+    }
+
+    @Override
+    public int compareTo(InformationLoss<?> other) {
+        if (!other.getClass().equals(this.getClass())) {
```
prasser: Is this really needed? Maybe this should be removed for efficiency reasons (but I don't know how often this method is actually used...)

src/main/org/deidentifier/arx/metric/v2/ILScore.java
```diff
@@ -0,0 +1,130 @@
...
+        }
+        BigFraction otherValue = ((ILScore)other).getValue();
+        return value.compareTo(otherValue) * -1;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) return true;
+        if (obj == null) return false;
+        if (getClass() != obj.getClass()) return false;
```
prasser: Is this really needed? Maybe this should be removed for efficiency reasons (but I don't know how often this method is actually used...)

src/main/org/deidentifier/arx/metric/v2/ILScore.java
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * Converter method.
+     *
+     * @param other
+     * @return
+     */
+    protected ILScore convert(InformationLoss<?> other) {
+        if (other == null) return null;
+        if (!other.getClass().equals(this.getClass())) {
```
prasser: Is this really needed? Maybe this should be removed for efficiency reasons (but I don't know how often this method is actually used...)

src/main/org/deidentifier/arx/ARXAnonymizer.java
RaffaelBild: I just filed an issue about this

src/main/org/deidentifier/arx/ARXLattice.java
RaffaelBild: In fact, it is not obvious if such a modification retains the expected differential privacy guarantees. In all relevant formulas/proofs, the search algorithm has the form of a function which returns a single generalization scheme. This modification would formally extend it to a function which returns a set of generalization schemes, and it would require further analytical investigation to determine possible effects on the privacy guarantees provided which could result from such an extension. Hence, at least for now, I would suggest to stay with the proposed solution which returns one generalization scheme and is formally safe.

PR #220, opened 2018-09-18
Resolve #219: Don't apply cleansing to headers.

### LINKED ISSUE #219: Specifying Data Types for CSV Files Crashes ARX
When specifying type information for a given column in a CSV ARX will automatically apply data cleansing to ensure types match throughout the column. It does this inside the CSV parser itself, via the iterators next() method. Unfortunately this is also the method used to extract headers from the CSV elsewhere in the code, which causes the data cleansing to be applied nulling out the headers, which then causes a NullPointerException deeper in ARX. Since the code appears to always assume headers are present, we should skip cleansing on the first row.

chawco:
This skips data cleansing on the first row to allow headers to be parsed elsewhere. This should resolve #219

### DIFF
--- a/src/main/org/deidentifier/arx/io/CSVDataInput.java
+++ b/src/main/org/deidentifier/arx/io/CSVDataInput.java
@@ -454,9 +454,10 @@ public class CSVDataInput {
         return new Iterator<String[]>() {
 
             // Next tuple
-            boolean   initialized = false;
-            CsvParser parser      = null;
-            String[]  next        = null;
+            boolean   initialized   = false;
+            boolean   parsedHeaders = false;
+            CsvParser parser        = null;
+            String[]  next          = null;
 
             @Override
             public boolean hasNext() {
@@ -479,7 +480,7 @@ public class CSVDataInput {
                 next = parser.parseNext();
                 
                 // Replace each non matching value with the special NULL string
-                if (cleansing) {
+                if (cleansing && parsedHeaders) {
 
                     if (result.length != datatypes.length) {
                         throw new IllegalArgumentException("More columns available in CSV file than data types specified");
@@ -491,6 +492,7 @@ public class CSVDataInput {
                         }
                     }
                 }
+                parsedHeaders = true;
                 return result;
             }
 


chawco:
Here's a simple test case:

```java
import java.io.*;
import java.nio.charset.Charset;
import org.deidentifier.arx.*;
import org.deidentifier.arx.criteria.KAnonymity;
import org.deidentifier.arx.io.CSVSyntax;

public class TestARXCSV {

  public static void main(String args[]) throws IOException, FileNotFoundException {
    // Generate our test CSV
    String csvContents = "string,int\nfoo,7";
    File file = new File("test.csv");
    PrintWriter csvOut = new PrintWriter(file);
    csvOut.write(csvContents);
    csvOut.flush();
    csvOut.close();

    Charset charset = Charset.forName("UTF-8");
    CSVSyntax csvSyntax = new CSVSyntax(',');

    // Set up our variables
    DataType<?>[] types = {DataType.STRING, DataType.STRING};
    Data data;
    DataDefinition def;
    ARXConfiguration config;
    ARXAnonymizer anonymizer;

    // Test1 is fine -- all strings, so nothing gets stripped
    data = Data.create(file, charset, csvSyntax, types);
    def = data.getDefinition();

    def.setAttributeType("string", AttributeType.INSENSITIVE_ATTRIBUTE);
    def.setAttributeType("int", AttributeType.QUASI_IDENTIFYING_ATTRIBUTE);

    config = ARXConfiguration.create();
    config.addPrivacyModel(new KAnonymity(1));
    anonymizer = new ARXAnonymizer();
    try {
      anonymizer.anonymize(data, config);
    } catch (Exception e) {
      System.out.println("Test1 failed!");
      e.printStackTrace();
      return;
    }

    System.out.println("Test1 passed");

    // Change the second column to be an INTEGER
    types[1] = DataType.INTEGER;

    // Test2 fails, because we stripped the header value "int" because
    // it's not a number
    data = Data.create(file, charset, csvSyntax, types);
    def = data.getDefinition();

    def.setAttributeType("string", AttributeType.INSENSITIVE_ATTRIBUTE);
    def.setAttributeType("int", AttributeType.QUASI_IDENTIFYING_ATTRIBUTE);

    config = ARXConfiguration.create();
    config.addPrivacyModel(new KAnonymity(1));
    anonymizer = new ARXAnonymizer();
    try {
      anonymizer.anonymize(data, config);
    } catch (Exception e) {
      System.out.println("Test2 failed:");
      e.printStackTrace();
      return;
    }

    System.out.println("Test2 passed");
  }
}
```

Running this yields:
```
trajan:arx-test craig$ javac -cp libarx-3.7.1.jar TestARXCSV.java
trajan:arx-test craig$ java -cp libarx-3.7.1.jar:. TestARXCSV
Test1 passed
Test2 failed:
java.lang.IndexOutOfBoundsException: Column index out of range: -1. Valid: 0 - 1
        at org.deidentifier.arx.DataHandle.checkColumn(DataHandle.java:840)
        at org.deidentifier.arx.DataHandleInput.getDistinctValues(DataHandleInput.java:254)
        at org.deidentifier.arx.DataHandle.getDistinctValues(DataHandle.java:156)
        at org.deidentifier.arx.DataDefinition.materializeHierarchies(DataDefinition.java:745)
        at org.deidentifier.arx.ARXAnonymizer.anonymize(ARXAnonymizer.java:199)
        at TestARXCSV.main(TestARXCSV.java:63)
trajan:arx-test craig$
```

Which is because `"int"` is not a number in the header, and so gets stripped out. Later on, ARX tries to rely on these column headers, but they've been stripped by the cleaning, and hence the nonsensical output above.

@prasser hopefully this helps demonstrate the issue sufficiently!

prasser:
Fix commited. Closing.

PR #223, opened 2018-09-25
Integrate data-dependent differential privacy into the GUI

RaffaelBild:


src/gui/org/deidentifier/arx/gui/Controller.java
```diff
@@ -597,9 +598,11 @@ public void actionMenuEditAnonymize() {
         }
         
         // Open configuration dialog
-        ModelAnonymizationConfiguration configuration = this.actionShowLocalAnonymizationDialog();
-        if (configuration == null) {
-            return;
+        if (!model.getInputConfig().containsCriterion(EDDifferentialPrivacy.class)) {
```
prasser: Why is the dialog not shown for the DP-Model? AFAICT, the dialog dynamically adjusts possible selections to the privacy model selected, so it may be a good idea to show it for DP as well... Or not?

RaffaelBild: I decided to not show the dialog because all the options would be disabled. Moreover, the notion of a "step" is overloaden. In the dialog, it refers to the heuristic search step limit, while the number of steps to use for the dp search can be specified in the dialog for configuring the dp privacy model.

As an alternative, it would be possible to remove the configuration of the number of steps to use from the dp privacy model configuration dialog, and to interpret the "Limited number of steps" as the number of search steps to use for the dp search whenever data-dependent dp is set. Which alternative would you prefer?

prasser: Definitely the second one!

RaffaelBild: Very well, I implemented that option

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogAnonymization.java
```diff
@@ -89,8 +94,17 @@ public DialogAnonymization(Shell shell, Model model) {
...
         this.configurationValid = true;
         
-        // Determine if optimal search is available
-        this.optimalSearchAvailable = model.getSolutionSpaceSize() <= model.getHeuristicSearchThreshold();
+        // Determine available search options
+        EDDifferentialPrivacy edp = model.getInputConfig().getCriterion(EDDifferentialPrivacy.class);
+        if (edp != null) {
+            this.optimalSearchAvailable = false;
+            this.heuristicSearchStepLimitAvailable = edp.isDataDependent();
+            this.heuristicSearchTimeLimitAvailable = false;
```
prasser: PrivacyModel.isHeuristicSearchSupported(); defaults to true
PrivacyModel.isHeuristicSearchWithTimeLimitSupported(); defaults to true
-> Overwrite for EDP

PR #224, opened 2018-10-16
add all deps for excel import

nartz:
Importing an excel file was broken, because it was missing some dependencies. This adds the necessary dependencies to properly use apache poi for reading excel.

Note: Poi is very picky about what components it needs:
http://poi.apache.org/components/index.html#components

### DIFF
--- a/.idea/libraries/Maven__commons_codec_commons_codec_1_5.xml
+++ b/.idea/libraries/Maven__commons_codec_commons_codec_1_5.xml
@@ -1,0 +1,13 @@
+<component name="libraryTable">
+  <library name="Maven: commons-codec:commons-codec:1.5">
+    <CLASSES>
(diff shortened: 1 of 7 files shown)


prasser:
Thanks! Are those libs also missing when ARX is built using Ant or are they just not defined in the maven pom?

nartz:
I just built the jars from master on osx-64 and importing an excel seemed to work fine, so this doesn't appear to be affecting the ant build.

prasser:
Thanks a lot for the quick feedback! I will pull this then.
-- 
Dr. Fabian Prasser

Institute of Medical Informatics, Statistics and Epidemiology
University Hospital rechts der Isar
Technical University of Munich

Grillparzerstr. 18
81675 Munich
Germany

Tel: +49 89 4140 4328
Fax: +49 89 4140 4850
Web: www.imse.med.tum.de

PR #247, opened 2019-02-01
First implementation of the Fisher-Yates Knuth-Yao permutation algorithm on column

giupardeb:
I implemented the first release of Fisher-Yates Knuth-Yao permutation algorithm and create a example file to test the permutation.

PR #253, opened 2019-03-16
 First implementation of the Rao-Sandelius permutation algorithm on column

giupardeb:
this is only the first implementation, I will work to improve the code further.

### DIFF
--- a/src/main/org/deidentifier/arx/masking/DataMaskingFunction.java
+++ b/src/main/org/deidentifier/arx/masking/DataMaskingFunction.java
@@ -167,17 +167,16 @@ public abstract class DataMaskingFunction implements Serializable {
 		private ArrayList<String> rs(int lengthColumn, ArrayList<String> column) {
 			
 			Random rand = new SecureRandom();
-			if(lengthColumn <= 1) {
+			if(lengthColumn <= 1)
 				return column;
-			}
+			
 			if(lengthColumn == 2) {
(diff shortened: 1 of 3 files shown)


prasser:
Thanks! Please take a look at the changes in your prior pull request (see https://github.com/arx-deidentifier/arx/pull/247). The function needs to be added in some further places and a MaskingFunctionDescription needs to be provided. Best, Fabian

giupardeb:
> Thanks! Please take a look at the changes in your prior pull request (see #247). The function needs to be added in some further places and a MaskingFunctionDescription needs to be provided. Best, Fabian

Thank you! I thought I did it..but I recheck better!

giupardeb:
I think that I have fixed all issues, but if I don't, please tell me and I will do all necessary things to fix it. Thanks!

src/main/org/deidentifier/arx/masking/DataMaskingFunction.java
```diff
@@ -161,9 +161,14 @@ private void raoSandelius(DataColumn column) {
 			
 			for (int i=0;i<lengthColumn;i++)
 				column.set(i, colout.get(i));
-			
 		}
 		
+		/**
+		 *
```
prasser: Please provide a description here

giupardeb: Done! :-)

PR #261, opened 2019-04-25
Replace the reliable variant of the exponential mechanism with a sufficiently robust floating-point implementation

RaffaelBild:


src/main/org/deidentifier/arx/algorithm/DataDependentEDDPAlgorithm.java
```diff
@@ -183,12 +171,15 @@ private long executeExponentialMechanism(Map<Long, ILScore> transformationIDToSc
...
         Long[] values = new Long[transformationIDToScore.size()];
-        BigFraction[] scores = new BigFraction[values.length];
+        double[] scores = new double[values.length];
         
         int i = 0;
         for (Entry<Long, ILScore> entry : transformationIDToScore.entrySet()) {
             values[i] = entry.getKey();
-            scores[i] = entry.getValue().getValue();
+            scores[i] = entry.getValue().getValue().doubleValue();
+            if (Double.isInfinite(scores[i]) || Double.isNaN(scores[i])) {
```
prasser: How can you be sure that no rounding is involved in the conversion to double? I can find nothing in that regard in the documentation. How do you know that NaN or +/-Infinity can be produced?

src/main/org/deidentifier/arx/dp/ExponentialMechanism.java
```diff
@@ -14,111 +14,147 @@
...
+ * 
  * This implementation assumes that all score values have been divided by the sensitivity of the respective score function.
  * 
- * Note: This implementations uses internal caches which may grow up to the size of distinct values
- * provided as scores to the method setDistribution during the lifetime of an instance.
+ * We point out that this implementation draws from a probability distribution which approximates the mathematically precise
+ * distribution of the exponential mechanism as a consequence of floating-point arithmetic, which could potentially affect
+ * the privacy guarantees provided. However, it can be shown that the resulting exceedance of the privacy parameter epsilon
+ * is at most in the order of log((1 + n * 10^{1-p}) / (1 - n * 10^{1-p})) where n is the size of the domain and p is
+ * the floating-point precision which can be specified by the user.
```
prasser: I see doubles being used below. Are you sure that we are not restricted by the precision of doubles here? What would that mean for the error described in the formula above?

src/main/org/deidentifier/arx/dp/ExponentialMechanism.java
```diff
@@ -128,120 +164,57 @@ public void setDistribution(T[] values, BigFraction[] scores) {
...
+        // Compute the probability for every value by calculating enumerator / divisor.
+        // Note that during this computation, the factor exp(-shift) is effectively being cancelled.
+        List<Pair<T, Double>> pmf = new ArrayList<>();
+        for (int i = 0; i < scores.length; ++i) {
+            BigDecimal enumerator = enumerators[i];
+            BigDecimal probability = enumerator.divide(divisor, mc);
+            // Add entry to the PMF.
+            // Note that BigDecimal.doubleValue() may yield infinity.
+            // In such a case, the constructor of the class EnumeratedDistribution will throw an exception.
+            pmf.add(new Pair<T, Double>(values[i], probability.doubleValue()));
```
prasser: How about rounding here?

PR #267, opened 2019-05-12
initial development of go back function from masked attribute

giupardeb:
there is still a lot of work to do, but before continuing development, I create this pull request because I am not sure if the position of_WriteFile_ method is correct. Can you help me? 
Thank you so much!

### DIFF
--- a/src/example/org/deidentifier/arx/examples/Example58.java
+++ b/src/example/org/deidentifier/arx/examples/Example58.java
@@ -98,6 +98,7 @@ public class Example58 extends Example {
 
         // Print info
         printResult(result, data);
+        System.out.println(result.getMapping(1));
 
         // Process results
         System.out.println(" - Transformed data:");
--- a/src/example/org/deidentifier/arx/examples/Example58.java
+++ b/src/example/org/deidentifier/arx/examples/Example58.java
@@ -98,7 +98,6 @@ public class Example58 extends Example {
 
         // Print info
         printResult(result, data);
-        System.out.println(result.getMapping(1));
 
         // Process results
         System.out.println(" - Transformed data:");
--- a/src/example/org/deidentifier/arx/examples/Example58.java
+++ b/src/example/org/deidentifier/arx/examples/Example58.java
@@ -95,6 +95,7 @@ public class Example58 extends Example {
         config.setSuppressionLimit(0d);
 
         ARXResult result = anonymizer.anonymize(data, config);
+        System.out.println(result.getPseudonymMappingTable().get("id"));
 
         // Print info
         printResult(result, data);
--- a/src/main/org/deidentifier/arx/ARXAnonymizer.java
+++ b/src/main/org/deidentifier/arx/ARXAnonymizer.java
@@ -747,6 +747,13 @@ public class ARXAnonymizer { // NO_UCD
         }
         return org.deidentifier.arx.framework.data.Data.createCombination(manager.getDataInput(), columns);
     }
+    
+    private void getMap(DataManager manager, DataDefinition definition) {
+    	DataColumn[] columns = manager.getDataMasking();
+    	for (DataColumn column : columns) {
+            definition.getMaskingFunction(column.getAttribute()).getMap();
+        }
+    }
 
     /**
      * Parses the settings provided by the given instance
--- a/src/main/org/deidentifier/arx/ARXAnonymizer.java
+++ b/src/main/org/deidentifier/arx/ARXAnonymizer.java
@@ -89,6 +89,9 @@ public class ARXAnonymizer { // NO_UCD
 
         /** Masked data */
         final org.deidentifier.arx.framework.data.Data maskedData;
+        
+        /** Clear data*/
+        final String[][] clearData;
 
         /**
          * Creates a new instance.
@@ -105,6 +108,7 @@ public class ARXAnonymizer { // NO_UCD
                final SolutionSpace solutionSpace,
                final DataManager manager,
                final org.deidentifier.arx.framework.data.Data maskedData,
+               final String[][] clearData,
                final AbstractAlgorithm algorithm,
                final long time,
                final boolean optimumFound) {
@@ -112,6 +116,7 @@ public class ARXAnonymizer { // NO_UCD
             this.solutionSpace = solutionSpace;
             this.manager = manager;
             this.maskedData = maskedData;
+            this.clearData = clearData;
             this.algorithm = algorithm;
             this.time = time;
             this.optimum = algorithm.getGlobalOptimum();
@@ -139,6 +144,7 @@ public class ARXAnonymizer { // NO_UCD
                                  handle.getRegistry(),
                                  this.manager,
                                  this.maskedData,
+                                 this.clearData,
                                  this.checker,
                                  handle.getDefinition(),
                                  config,
@@ -722,7 +728,9 @@ public class ARXAnonymizer { // NO_UCD
         long time = System.currentTimeMillis();
         
         // Perform data masking
-        org.deidentifier.arx.framework.data.Data maskedData = mask(manager, definition); 
+        org.deidentifier.arx.framework.data.Data maskedData = mask(manager, definition);
+        
+        String[][] clearData = manager.getDataInput().getDictionary().getMapping();
 
         // Execute search algorithm
         boolean optimumFound = algorithm.traverse();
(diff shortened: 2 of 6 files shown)


giupardeb:
Another question, where can I save the mapping file ?
Thanks!

prasser:
Thanks. This should be solved a bit differently. The mapping table should be made accessible to the user, who can then store or further process it. For this purpose, the table should be associated with the instance of ARXResult and a method for access should be offered, e.g.:

ARXResult.getPseudonymMappingTable()

It would also be nice, if this function returned not just a Map<> but maybe a special class (could, e.g. be an instance of DataHandle).

giupardeb:
Hi Fabian, I'm working a lot on the implementation of this feature. My goal is to have the map inside the anonymize function (ARXAnonymizer.java) in such a way as to create an instance of the Result class that contains the mapping table for each column. Is what I've done so far correct? Also in the getMap function, how can I keep the map I get for each column? Thanks a lot for your time!

giupardeb:
I thought of using the data inside the manager object to retrieve the initial data. What do you think about it?

prasser:
No, I wouldn't do that. You already have access to input data by using Data.getHandle(). This additional method is not needed. So you don't even need to add anything to ARX to give the user access to the mapping. The mapping can be accessed by using input.getValue(row, col) and output.getValue(row, col). However, it could still be a good idea to provide the mapping to the user via ARX Result.getMapping(col) or something similar and attaching it to DataManager is a good idea.

giupardeb:
Thanks a lot for your suggestion! There are a lot work to do, but I hope that this commit is the right way.

giupardeb:
any news? @prasser 
Thank you so much for your time!

prasser:
You need to decide on a clear implementation strategy. Either you want to (1) implement a generic getMapping() method (independent of masking) or (2) specifically create the mapping during the masking operation. Currently you have a mixture of (1) and (2). If you want to implement (1) you should move this onto the level of DataHandle (no changes in DataManager etc.), if you want to implement (2) you should generate the mapping on the level of the masking function.

giupardeb:
I want to implement the (2) strategy. So I have to develop the getMapping function for each class that extends DataMaskingFunction and then I have to develop the function inside the ARXResult class in order to allow the user to store map or further process it. Is that correct?

prasser:
No, I would suggest to provide a special method within the base class for MaskingFunctions that must then be used by all implementations and create the mapping in the background. Something like that. Btw: why is (2) better than (1)?

giupardeb:
I consulted with my professor and we believe option (2) is the more appropriate. We imagine reidentification to depend on the specific pseudonymization technique. For example, reidentification following permutation will differ from reidentification following noise addition. Hence, each pseudonymization technique should come with its reidentification algorithm

giupardeb:
with this commit, I developed a function that allow me to create a mapping. Now, in according to your previous message, I will develop a ARXResult.getPseudonymMappingTable() function into a ARXResult class. is it right?

giupardeb:
Hi Fabian, I worked to add a getPseudonymMappingTable inside the ARXResult class. Is this the correct way? I'm doing my best to write a correct solution

prasser:
Quick feedback:

(1) Please don't store the mapping in class MaskingFunction. It needs to be stored in DataManager.

(2) I still don't understand why you are implementing this. What is it needed for? The following method (which can easily be implemented without modifying ARX) does exactly the same as your current implementation (only works on the `msk` branch):

`
    /**
     * Return mapping for masked attributes
     * @param input
     * @param output
     * @return
     */
    private static Map<String, Map<String,String>> getMappingTable(DataHandle input, DataHandle output) {

        // Prepare
        DataDefinition definition = output.getDefinition();
        Map<String, Map<String,String>> result = new HashMap<>();

        // Check if handles match
        if (input.getNumRows() != output.getNumRows()) {
            throw new IllegalArgumentException("Handles don't match");
        }
        
        // For each masked attribute
        for (String attribute : definition.getIdentifyingAttributesWithMasking()) {
        
            // Get index
            int column = output.getColumnIndexOf(attribute);
            
            // Check if handles match
            if (input.getColumnIndexOf(attribute) != column) {
                throw new IllegalArgumentException("Handles don't match");
            }
            
            // Create map for this column and store
            Map<String, String> map = new HashMap<>();
            result.put(attribute, map);
            
            // Put values in map
            for (int row = 0; row < output.getNumRows(); row++) {
                map.put(output.getValue(row, column), input.getValue(row, column));
            }
        }
        
        // Done
        return result;
    }`

You need to think about the concrete problem that you want to solve before implementing anything. And when you have defined your concrete problem, you need to think about whether this really needs to be implemented by modifying ARX.

giupardeb:
Dear Fabian, thank you for your feedback.
I implemented the creation of the mapping table at the level of masking function,  as per your suggestion:

> if you want to implement (2) you should generate the mapping on the level of the masking function.

Moreover I also implemented  getPseudonymMappingTable function inside ARXResult as per your suggestion:

> The mapping table should be made accessible to the user, who can then store or further process it. For this purpose, the table should be associated with the instance of ARXResult and a method for access should be offered, e.g.:
> 
> ARXResult.getPseudonymMappingTable()

Maybe, I think that there are misunderstanding.

prasser:
Yes, there are misunderstandings. For example, I also commented the following on May 19th:

> No, I wouldn't do that. You already have access to input data by using Data.getHandle(). This 
> additional method is not needed. So you don't even need to add anything to ARX to give the user 
> access to the mapping. The mapping can be accessed by using input.getValue(row, col) and 
> output.getValue(row, col).

The code I provided above implements exactly that. It is very simple and it does the job. I also wrote that:

> However, it **could** still be a good idea to provide the mapping to the user via ARX 
> Result.getMapping(col) or something similar and attaching it to DataManager is a good idea.

As I said, **maybe** it's a good idea to provide the mapping to users. However, you need to find reasons for this and why the code that I provided above isn't sufficient. If you are doing this as a student project, you can add whatever you want to your local fork of ARX. However, I can only merge code into the main repository, which is well justified. In this case, I don't even understand which problem you are trying to solve. Just as a friendly suggestion: you should really try to develop a clear understanding of your requirements and goals before you start implementing anything.

giupardeb:
Hi Fabian, the problem I have in mind is that of a trusted superuser who pseudoanonymises a table, eg by Random Alphanumeric String function, but also wants to be able to recover the original table *without* storing the original table for security reasons. A solution would then be to produce two tables: table1 with the pseudoanonymised values, and table2 with the tuple that reconstructs the original table (e.g. id 1, 2, 3, 4, 5, 6... of the pseudoanonymised table). The superuser will then give table1 to user1 and table2 to user2 and finally delete the original table. It is clear that any attack to individual users and respective tables will protect the data. This is the scenario I am targeting. Incidentally, it is clear that the contents of table2 depend on the specific pseudonymisation technique. My latest commits is my attempt to build also table2.

prasser:
OK, I clearly understand that. The next questions are: why is the method that I provided above not sufficient? What specific functionality is needed that is not covered by this implementation? Do you want to add this to the API only, or to the GUI? If also to the GUI, how should the workflow look like?

giupardeb:
> why is the method that I provided above not sufficient?
I believe that your code implement a generic getMappingTable() method (independent of masking), while I would like to create the table during the masking operation (dependent of masking).

I thought that it is better to implement as API at the moment, because for me it's more easier.

giupardeb:
Hi @prasser , I was wondering if you had a chance to consider my message on why I think your code would not address the scenario I outlined earlier to you.  Time's elapsing quickly and I need to close this project asap, so may I kindly ask you to let me know if you think ARX would benefit from some implementation on my idea of re-identification. If yes, I'd be more than happy to hammer my implementation down as appropriate.

prasser:
Hi, sorry, I did not see your reply.

I'm not convinced by your arguments, unfortunately. In ARX, we aim to support user requirements. From this perspective, my implementation does exactly the same as yours. Moreover, it has the benefit that it is much simpler and much less invasive (less code, less classes changed etc.). I see two things that would need to be done:

(1) Integrate the method into ARX so that it becomes available to users. It should probably be added to DataHandleOutput, so that users can call handle.getMapping(). This would be very convenient.

(2) It would be great if this could be added to the GUI. For example, it should become possible to export the mapping together with the output file. That's why I asked for the concrete workflow that you are planning to support.

For your student project, it should not be required to get your code merged into ARX. So you can implement whatever you want and describe it in your project report.  If you want to get your code merged you should do (1) and maybe (2). Be aware, however, that it may still be difficult to get it merged soon, as we strive for high code quality and several iterations may be needed.

Good luck!

giupardeb:
Hi Fabian, Thank you so much for your time. 

Good Luck to you too!

PR #271, opened 2019-05-22
Maven core

dfirman:
Added a maven build profile that will package the code without any GUI components.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -655,6 +655,10 @@
                     <family>mac</family>
                     <arch>x86_64</arch>
                 </os>
+                <property>
+                    <name>core</name>
+                    <value>!true</value>
+                </property>
             </activation>
             <properties>
                 <envClassifier>osx-64</envClassifier>
@@ -681,6 +685,10 @@
                     <family>Windows</family>
                     <arch>x86_64</arch>
                 </os>
+                <property>
+                    <name>core</name>
+                    <value>!true</value>
+                </property>
             </activation>
             <properties>
                 <envClassifier>win-64</envClassifier>
@@ -706,6 +714,10 @@
                     <family>Windows</family>
                     <arch>x86</arch>
                 </os>
+                <property>
+                    <name>core</name>
+                    <value>!true</value>
+                </property>
             </activation>
             <properties>
                 <envClassifier>win</envClassifier>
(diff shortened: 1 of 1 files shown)


prasser:
Thanks for your interest in ARX and for your contribution!

However, I have some questions regarding this PR:

(1) Why is src/example added? I don't think that this folder should be part of a code package.
(2) Why are .properties files added from src/gui? They are not needed in the core of ARX.
(3) Why is TestResourceBundle.java excluded? AFAICS tests are not included in the package.

In addition, please don't move Resources.java. Moreover, the way it has been implemented here, probably breaks the GUI-builds.

dfirman:
To answer your questions,

(1) You are correct, src/example should not be included. I will change that.
(2) .properties is included from src/gui because that is where the version number is. In AbstractAnonymizationTest.java the test() function uses the version number from those properties.
(3) When running mvn package tests are run by default and must pass in order for jar to be produced.

Overall I think most of these issues are because the version number is maintained within src/gui but the unit tests depend on it. One solution could be to just remove that line from AbstractAnonymizationTest.java, however TestResourceBundle.java would still fail to compile without the src/gui resources. I did not want to remove the version number from messages.properties because that would break other code. I also did not want to copy that number because that would require maintaining the version number in two places. Let me know what you think about this.
Is there an issue with having a Resources.java in both src/gui and src/main? The GUI version still builds from my tests.

prasser:
Thanks for the quick feedback! This  makes sense. I'll try to come up with a solution and let you know what to do ASAP.

prasser:
I have just committed some changes that should fix these issues. Will this work for you?

BTW: I took the opportunity to bump the version number to 3.8.0

dfirman:
I think these changes will work, Thanks! However there seems to be an issue with the current master maven build. I will create a new issue for it and continue this branch when it is resolved.

dfirman:
Ok I have updated the branch. The only think left now is src/examples. Currently it is still in the build profile because the unit test TestExamples.java depends on it. I know it is possible to skip running specific tests in maven, but all the tests must still compile.
The two solutions I see are to 
 (1) include src/example in the build profile or 
 (2) move examples into the tests if they are only used for testing

Let me know what you think, or if you have any other solutions.

prasser:
Looks good to me. @nartz any objections?

nartz:
Lgtm. I would add a few lines to the readme outlining the standard maven one-liner for building core, versus building with GUI.

prasser:
OK, I'll merge this now. Would be great if @dfirman or @nartz could propose a few lines to add to the readme.

PR #275, opened 2019-06-03
Shade dependencies

dfirman:
I ran into an a dependency conflict with a project. Shading some common dependencies fixed the issue.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -93,40 +93,6 @@
                                     <mainClass>org.deidentifier.arx.gui.Main</mainClass>
                                 </transformer>
                             </transformers>
-                            <relocations>
-                                <relocation>
-                                    <pattern>org.apache</pattern>
-                                    <shadedPattern>org.shaded.apache</shadedPattern>
-                                </relocation>
-                                <relocation>
-                                    <pattern>com.google</pattern>
-                                    <shadedPattern>com.shaded.google</shadedPattern>
-                                </relocation>
-                                <relocation>
-                                    <pattern>com.googlecode</pattern>
-                                    <shadedPattern>com.shaded.googlecode</shadedPattern>
-                                </relocation>
-                                <relocation>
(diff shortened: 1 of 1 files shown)


nartz:
I might question if this is should be part of master, since the shading seems to be part of your project specific configuration, no? Can you think of a reason otherwise?

dfirman:
These changes are not project specific. arx is using specific versions of a lot of common dependencies, for example, `com.google.guava`. This means anyone that wants to use arx as a dependency could run into the same issues with the transitive dependencies from arx. 
I will remove the shading on the non-core profiles because those should never be a dependancy. I do think that dependencies should be shaded for the core profile in order to use arx as a dependency. 
However, If you do want arx to be a dependancy for other projects than the shading doesn't matter.

nartz:
I could see this going either way I guess, but I'm still leaning to the fact that this could belong in a private fork.

The reason is, dependency conflicts will be different for each project. While these might be a reasonable set of defaults to shade that you've provided, other developers may need none of these, or more of them; without shading all dependencies its difficult to know.

Also note that shading comes with some of its own problems. I haven't personally encountered all of these issues, but here's a small discussion, note what was said about the elastic search team moving away from shading. 

https://softwareengineering.stackexchange.com/questions/297276/what-is-a-shaded-java-dependency
https://www.elastic.co/blog/to-shade-or-not-to-shade

Just my 2 cents. @prasser might have different thoughts.

prasser:
I had the same thoughts. I think we should either shade all external dependencies or none of them. So, it boils down to whether or not we want shading at all. I'm not a very frequent Maven user, so I'm unsure about what's the right thing to do here...

dfirman:
There is no good way to shade all dependencies, which is why I only chose these "common" ones. Since this project is open source I suppose It is fine to let users figure out dependencies issues if they arise for them. 
Thanks for the input, I will close this PR.

PR #281, opened 2019-06-19
Add parameter checks

RaffaelBild:


nartz:
just fyi - wondering why this single merge contains 27 other commits? have you merged upstream master into your branch (recently)? If you care to get rid of those, you could squash them or rebase on master or otherwise. Not sure if @prasser  cares about such things...

prasser:
@nartz I do care, but in this case I can simply squash and merge. Is there anything wrong with that?

PR #328, opened 2020-09-11
How to handle missing and invalid data?

nesokreso:
Add possibility to have empty strings in csv for dataType DATE

### DIFF
--- a/src/main/org/deidentifier/arx/DataType.java
+++ b/src/main/org/deidentifier/arx/DataType.java
@@ -319,7 +319,7 @@ public abstract class DataType<T> implements Serializable, Comparator<T> { // NO
         
         @Override
         public Date parse(String s) {
-            if(s.length() == NULL_VALUE.length() && s.toUpperCase().equals(NULL_VALUE) || s.equals("")) {
+            if(s.length() == NULL_VALUE.length() && s.toUpperCase().equals(NULL_VALUE) || s.isEmpty()) {
                 return null;
             }
         	try {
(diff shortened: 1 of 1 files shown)


arx-deidentifier:
Thanks for this proposal. What happens when an empty string is provided as a date without this patch? What happens when empty steings are provided for variables of other data types (e.g. Interner, decimal)?

nesokreso:
It throws an exception because an empty String can not be parsed to Date, Integer, Decimal e.g. So with this check it will be converted to a NULL value an the DataSource can be anonymized.

prasser:
I don't think that this is needed, as ARX supports data cleansing during loading. Try this:

Without cleansing:
source.addColumn("Age", DataType.INTEGER);

With cleansing:
source.addColumn("Age", DataType.INTEGER, true);

Does that solve your issue?

nesokreso:
Ok with the cleansing parameter my example runs without exceptions. 
In my opinion it would be nicer, when this is also checked in the parse function. But if you don't want this commit i would close this pull request.

prasser:
IMHO, there are good reasons for _not_ including this in the parse function. An empty string is just one example of an infinite number of invalid values that cannot be parsed. Hence, I see no meaningful reason for including this one example into the parse function, but not others. An addition, how invalid values should be treated is generally a difficult problem (just turning every invalid value into NULL might be problematic, because users might for example need to distinguish between invalid values and missing values). Hence, it makes sense to leave data cleansing to the users and perform it only if explicitly requested. Will close this now,

PR #337, opened 2020-10-15
Fixed DP-Test

srcds:


prasser:
Thanks! Please clarify that this doesn't break other things (cf. SearchStepSemantics.EXPANSIONS vs CHECKS).

srcds:
The mustStop() statement was introduced during refactoring the AbstractAlgorithm class. Before this refactoring took place the inner-loop did not have any mechanism to prematurely terminate. Therefore, it can be assumed that the mustStop-statement in the inner loop was added by mistake and that removing it will restore the original workflow. However, instead of completly removing it, one could also consider to simply replace its "return false" with a "break" command. Using "break" the Unit-tests also pass.

prasser:
Thanks. A few comments:

(1) I don't think that adding this statement was a mistake. I think that the underlying idea was that the algorithm can also be terminated via a time limit (which was not possible before). However, the DP-algorithm always requires (a) a step limit (with semantics "Expansions") or (b) it performs just one check. So please check whether mustStop() checks the step limit *and* the time limit, or not. If it only checks one of both, depending on what has been configured, the statement can be removed. Even if this is not the case, we may decide to drop support for the time limit for the DP algorithm at this time to proceed in the development cycle. However, we then need to make sure that the fact that both types of limits cannot be used simultaineously is also reflected in the GUI.

(2) Simply replacing "return false" with "break" is dangerous! We would need to make sure that this doesn't break the privacy guarantees promised by the DP algorithm.

srcds:
Thanks for your feedback!

(1) mustStop() checks both, time and steps. However, even by removing the mustStop() statement in the inner loop I would still consider it possible to use both, time and steps, simultaneously as the outer loop still contains its own mustStop() statement. Of course both limits may be slightly exceeded depending on the complexity of the inner loop

(2) Ok, as the inner loop iterates over a list of possible transformations calculating their "ILScore" I thought it is ok to preemtivley interrupt the loop without putting the privavy guarantees at risk. However, I certainly do not have a profound understaning of the algorithm.

prasser:
Ok, thanks. Response to (1) got me convinced. Pulling!

PR #342, opened 2020-12-30
Modify bug

maricolm:
Hello, the code content has been modified again according to your requirements. Please check it and give your comments。3ks


prasser:
Thanks for the update, but this is not what I meant. I suggest the following way forward:

(1) Add your changes to arx.iml and pom.xml as well as the new example to your _master_ branch.
(2) Recreate a pull request from your master against the new branch arx/jdbc
(3) I will then pull this and modify the code in this branch myself before merging it into master.

One last point: please also comment on the source for the if-statement to determine the product type and the source for the syntax used to create product-specific SQL statements.

Thanks again!

PR #357, opened 2021-05-12
Fix issue on gui

iylee71:
When generating hierarchy, stack overflow occurs if 'Aggregate Function' that have parameter is selected in 'General Setting'.


prasser:
Thanks!

PR #359, opened 2021-06-24
Fixed issue with hierarchy import - closes #358

### LINKED ISSUE #358: [BUG] Unable to replace hierarchies after using the hierarchy wizard
Once the hierarchy wizard was used to create a hierarchy for an attribute, this hierarchy cannot be properly replaced with an imported hierarchy. Although importing the new hierarchy does not result in any errors and the imported hierarchy is correctly displayed in the GUI, the anonymization process still utilizes the old hierarchy which was created by the wizard. 

Tested on ARX 3.9.0 and on ARX 3.8.0 (Windows).

Steps for reproduction:

1.  Create new project and import data, E.g.:
> Attribute_1
A
B
C
D

2. Create a hierarchy for Attribute_1 using the wizard. I used the “Use masking” option and clicked through the wizard resulting in the following hierarchy: 
>A;*
B;*
C;*
D;*

3. Import a new hierarchy for Attribute_1. E.g. the following (manually created using a notepad):
> A;low;*
B;low;*
C;high;*
D;high;*

4. Set 2-Anonymity; Make Attribute_1 a Quasi-identifier; Anonymize.

Looking on the result reveals, that ARX won’t use the imported hierarchy but the one created with the wizard.

The .deid project file of this example as well as the raw dummy data and dummy hierarchy is attached in the .zip file.
[ExampleHierarchyBug.zip](https://github.com/arx-deidentifier/arx/files/6694655/ExampleHierarchyBug.zip)

srcds:


prasser:
Thanks, looks good! Please reference the issue solved in the PR title. Then GitHub will link both to each other, which is quite handy. Then it can be merged.

PR #368, opened 2021-11-29
Adding a bash script to handle removing swt libraries for other systems automatically 

idhamari:
Running this script will remove SWT for other systems e.g. if you are in windows it will remove libs for mac and linux. 
The script is already tested on Linux Ubuntu and Windows with gitbash.
The script should be run before opening arx in eclipse.


prasser:
Thanks. There are multiple problems with this PR:

(1) It includes unrelated changes.

See: https://github.com/arx-deidentifier/arx/pull/368/files

Unrelated changes: 
- There is an unneeded file: runMeFirst.sh.bak 
- There are changes in the following files:
src/gui/org/deidentifier/arx/gui/Controller.java
src/gui/org/deidentifier/arx/gui/view/impl/MainWindow.java

(2) The change is generally not needed. It is not required to delete the swt libraries if the project is configured corrently in Eclipse.

Won't merge.

Issue #366, opened 2021-11-26
Bug in hierarchy wizard for dates when using „Quarter of year”


srcds:
Selecting the “Quarter of year” option when using the hierarchy wizard for dates results in intervals of size 4. The first quarter ranges from ]0, 4] and includes the months 1-3. The second quarter ranges from ]4, 8] and includes the months 4-7, which should not be the case.

![grafik](https://user-images.githubusercontent.com/9593020/143660595-762c54c9-1641-4030-8e0e-18a5b2ea1d54.png)

prasser:
Assigned to @srcds as this is a good oppportunity to better get to know ARX.

idhamari:
**To Reproduce:** 

1. extract the project in this [compressed file sampleData.zip](https://github.com/arx-deidentifier/arx/files/7626200/sampleData.zip). Note: data source is [kaggle](https://www.kaggle.com/shivamb/football-deaths/version/1).
2. Open ARX then File>Open Project> arx/data/iaDatesSample.deid
2. Then Edit>Create hierarchy>use dates (for dates)>Quarter of year

Sample output:

![Screenshot from 2021-11-30 14-54-51](https://user-images.githubusercontent.com/28802516/144060127-86c5b803-13a3-4620-bb68-53be4b600621.png)

The desired output should group the months to be something like:

                 Months     Month indices 
                  1-3             0-2
                  4-6             3-5   
                  7-9             6-8
                  10-12           9-11

**Proposed Solution:**

The code responsible for this behavior can be found [here] (https://github.com/arx-deidentifier/arx/blob/204214fb1a4c56da0fcc85866b69d8bf60452ccd/src/main/org/deidentifier/arx/aggregates/HierarchyBuilderDate.java#L537)                  


            int year = Integer.valueOf(sdf.format(date));
            int lower = (year / _range) * _range;
            int upper = ((year / _range) + 1) * _range;
            return "[" + lower + ", " + upper + "[";

A good solution would be modifying the above equations to produce the correct output. For now, I solved the problem  by introducing a mapping function

	    /**
	     * Map a year to a range  to group months 
	     * @param input
	     * @param granularity
	     * @return
	     */
	    private String year_range_map(String month, int range ) {
	      Map<String , String> month_map = new HashMap < String, String>();
	      int lower = 1; 
	      for (int i=1; i<=12; i++) {
[...]

PR #371, opened 2021-12-13
Fix a bug in date hierarchy month/year

idhamari:
Fix ARX issue [366](https://github.com/arx-deidentifier/arx/issues/366).

### DIFF
--- a/src/main/org/deidentifier/arx/aggregates/HierarchyBuilderDate.java
+++ b/src/main/org/deidentifier/arx/aggregates/HierarchyBuilderDate.java
@@ -188,7 +188,7 @@ public class HierarchyBuilderDate extends HierarchyBuilder<Date> implements Seri
         /**  Granularity */
         WEEK("W"),
         /**  Granularity */
-        QUARTER("MM", 4),
+        QUARTER("MM", 3),
         /**  Granularity */
         YEAR("yyyy"),
         /**  Granularity */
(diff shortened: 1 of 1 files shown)


src/main/org/deidentifier/arx/aggregates/HierarchyBuilderDate.java
```diff
@@ -534,10 +534,11 @@ private String generalize(String input, Granularity granularity) {
...
             return sdf.format(date);
         } else {
-            int year = Integer.valueOf(sdf.format(date));
-            int lower = (year / _range) * _range;
-            int upper = ((year / _range) + 1) * _range;
-            return "[" + lower + ", " + upper + "[";
+            int dateUnit = Integer.valueOf(sdf.format(date));
+            int lower    = Integer.valueOf((dateUnit - 1) / (_range))  * (_range) + 1;
+            int upper    = lower + _range - 1;
+            String outputDate = "["+lower+"-"+upper+"]";
```
prasser, line 540: There are blanks missing before and after "+" in this line.

prasser (changes requested):
Looks good apart from minor formatting issues. One question: have you checked wether this also works with Millenia, e.g.? Or are you very confident that the code in lines 537-540 is correct for all cases (not only quarters)?

src/main/org/deidentifier/arx/aggregates/HierarchyBuilderDate.java
idhamari: > There are blanks missing before and after "+" in this line.

Do you mean line 540? I kept the old line format. If you like I can update it as well. 

> have you checked wether this also works with Millenia, e.g.? Or are you very confident that the code in lines 537-540 is correct for all cases (not only quarters)?

I already tested it on the Month/Year, Quarter a year, Decade, Century, and Millennium cases and it seems to work.

prasser: Yes, I mean line 540. And no: you did not keep the old line format ;)

image

PS: Please don't close PRs when changes are requested, but push commit to your branch to adress them.

PR #372, opened 2021-12-13
Fix a bug in date hierarchy month/year

idhamari:
Fix Arx [issue 366](https://github.com/arx-deidentifier/arx/issues/366).

### DIFF
--- a/src/main/org/deidentifier/arx/aggregates/HierarchyBuilderDate.java
+++ b/src/main/org/deidentifier/arx/aggregates/HierarchyBuilderDate.java
@@ -188,7 +188,7 @@ public class HierarchyBuilderDate extends HierarchyBuilder<Date> implements Seri
         /**  Granularity */
         WEEK("W"),
         /**  Granularity */
-        QUARTER("MM", 4),
+        QUARTER("MM", 3),
         /**  Granularity */
         YEAR("yyyy"),
         /**  Granularity */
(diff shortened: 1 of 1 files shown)


src/main/org/deidentifier/arx/aggregates/HierarchyBuilderDate.java
```diff
@@ -534,10 +534,11 @@ private String generalize(String input, Granularity granularity) {
...
             return sdf.format(date);
         } else {
-            int year = Integer.valueOf(sdf.format(date));
-            int lower = (year / _range) * _range;
-            int upper = ((year / _range) + 1) * _range;
-            return "[" + lower + ", " + upper + "[";
+            int dateUnit = Integer.valueOf(sdf.format(date));
+            int lower    = Integer.valueOf((dateUnit - 1) / (_range))  * (_range) + 1;
+            int upper    = lower + _range - 1;
+            String outputDate = "[" + lower + "-" + upper + "]";
```
prasser, line 540: The return format [1-3] breaks with the old format which would be [1, 4[. Please adjust. Line 540 can remain unchanged. line 539 should be rewritten as upper    = lower + _range;

prasser (changes requested):

idhamari:
@prasser code is updated!

prasser:
Just for the record: this change as well as as PR #372 is not correct. The problem is that year (basis for decade, millennium etc.) is 0-based but month (basis for quarter) is 1-based. Hence, different offsets are needed when calculating intervals and the current code is wrong for decades, millennia etc. This has been fixed in the following commit: https://github.com/arx-deidentifier/arx/commit/e1d20d68e1fcabb53fa127286ff1c3ea91c6717b

Issue #364, opened 2021-11-26
Add "Quarter/Year" to hierarchy wizard for dates


srcds:
When creating a hierarchy for dates using the option "Quarter of year", only the quarter but not the year is listed in the resulting hierarchy. As this also applies for "Month of year" (where only the month is listed), I guess this is not a bug. However, it would be useful to introduce the additional option “Quarter/Year” (similar to “Month/Year”), where both, the quarter, and the year, are considered in the generated hierarchy.

prasser:
Assigned to @srcds as this is a good oppportunity to better get to know ARX.

idhamari:
The code can be found in this [commit](https://github.com/ia-dev/arx/commit/145e39a26b3e0e99623f1ebc7cd3c6b983ed165f).

prasser:
Resolved

PR #373, opened 2021-12-13
Add option Quarter/Year

idhamari:
Fix ARX [issue 364](https://github.com/arx-deidentifier/arx/issues/364)


prasser:
I took the time today, to check this. This change as well as as PR #372 is not correct. The problem is that year (basis for decade, millennium etc.) is 0-based but month (basis for quarter) is 1-based. Hence, different offsets are needed when calculating intervals and the current code is wrong for decades, millennia etc.

What I now did is to use DateTimeFormatter instead of SimpleDateFormatter which also supports quarters and use the interval only for years.

See here: https://github.com/arx-deidentifier/arx/commit/e1d20d68e1fcabb53fa127286ff1c3ea91c6717b

idhamari:
@prasser  

> The problem is that year (basis for decade, millennium etc.) is 0-based but month (basis for quarter) is 1-based. Hence, different offsets are needed when calculating intervals and the current code is wrong for decades, millennia etc.

interesting! 

> What I now did is to use DateTimeFormatter instead of SimpleDateFormatter which also supports quarters and use the interval only for years.

Thanks for the correction.

Issue #367, opened 2021-11-28
Remove dynamic captions of checkboxes in settings dialog


srcds:
In the settings dialog there are several binary options which are controlled via ticking or unticking checkboxes. The caption right to the checkbox changes between “Yes”/”No” depending on whether the option is selected or not. I find this highly confusing.

I wanted to configure a project in such a way that all attributes of suppressed records were removed. Looking at the Transformation tab, I thought this behavior was already configured:
![grafik](https://user-images.githubusercontent.com/9593020/143791220-4b1475a6-69e1-414b-aa89-037985fb0202.png)

In my mind, the default for the second and third option was “Yes” and the non-default “No” must be selected manually by ticking it. As the anonymization was not removing all attributes, I revisited the options dialog and realized that the captions of the checkboxes change depending on whether they are ticked:
![grafik](https://user-images.githubusercontent.com/9593020/143791229-6d267169-d186-4a69-a996-812372433e80.png)
The problem: a “Unticked No” can be interpreted the same way as a “Ticked Yes”. I would strongly suggest changing the caption of the checkboxes to a static “Yes” or remove the caption completely. This change is easy to implement and I can't think of any disadvantages.

idhamari:
Captions are removed. I will create a pull request for this soon when the solution is approved.

 ![Screenshot from 2021-12-03 08-48-30](https://user-images.githubusercontent.com/28802516/144566325-3494fe07-dfdb-4b06-b089-52f6d68e14d3.png).

Some details: 

The code for this part can be found in [createTabTransformation](https://github.com/arx-deidentifier/arx/blob/204214fb1a4c56da0fcc85866b69d8bf60452ccd/src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogProperties.java#L303) function. This function uses import de.linearbits.preferences.PreferencesDialog; which is part of  [swtpreferences](https://github.com/arx-deidentifier/arx/blob/204214fb1a4c56da0fcc85866b69d8bf60452ccd/lib/ant/swtpreferences/swtpreferences-0.0.1.jar) library. The source code for this library is not provided in arx, it is provided [here](https://github.com/prasser/swtpreferences).

idhamari:
This issue is fixed in this [commit](https://github.com/iaBIH/swtpreferences/commit/c5aa9d012cb58e07b5794e0435fa50a6d79aad17)

prasser:
Fix commit

PR #374, opened 2021-12-16
fix issue 367: update setting and libraries

idhamari:
Handling [Remove dynamic captions of checkboxes in settings dialog #367](https://github.com/arx-deidentifier/arx/issues/367)


prasser:
There are more boolean options defined in the settings. Please modify them all to not show the label.

prasser:
Thanks. Now we have a follow-up issue. Line 294, 334 and 340 change the current behavior, as they introduce a default that wasnt there before. Can you please check what the ARX default for these options is? You can do this by following the code called in getValue() to see what is returned in the default case.

idhamari:
> Thanks. Now we have a follow-up issue. Line 294, 334 and 340 change the current behavior, as they introduce a default that wasnt there before. Can you please check what the ARX default for these options is? You can do this by following the code called in getValue() to see what is returned in the default case.

Thanks for the correction. I didn't check that and expected it to be false but it was actually true. The code is corrected.

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogProperties.java
```diff
@@ -291,7 +291,7 @@ private void createTabSearch(PreferencesDialog window) {
 
         window.addGroup(Resources.getMessage("DialogProperties.10")); //$NON-NLS-1$
 
-        window.addPreference(new PreferenceBoolean(Resources.getMessage("PropertyDialog.44")) { //$NON-NLS-1$
+        window.addPreference(new PreferenceBoolean(Resources.getMessage("PropertyDialog.44"), true, true) { //$NON-NLS-1$
```
prasser, line 294: How did you come to the conclusion that "true" should be the default here?

prasser: I think that it should be "false"...

prasser (changes requested):
See comment

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogProperties.java
idhamari: > How did you come to the conclusion that "true" should be the default here?

I called the function in swtprefernce example and got Yes:

         window.addPreference(new PreferenceBoolean("BooleanPreference with default") {
            protected Boolean getValue() { return true; }
            protected void setValue(Object t) { /*(Boolean)t*/}
        });

prasser: Hmm. I don't understand. In the screenshot from you code you can see, that swtpreferences doesn't offer the option to set the default for ""BooleanPreference with default" (blue arrow is grayed out). So you either 

(1) need to introduce a new constructor into swtpreferences to set hideLabel without setting a default (because currently there is no default set in ARX for the preferences under discussion). That would be the 100% downwards-compatible solution. Feel free to implement it (add new constructor to PreferenceBoolan and update swtprefereces again).

(2) Introduce a default. But it has to be the correct default. And this has nothing to do with swtpreferences or the example. It has to do with ARX. Every preference exposes one configuration option of ARX.

Let us take a look at this specific example:

```java
window.addPreference(new PreferenceBoolean(Resources.getMessage("PropertyDialog.44"), true, true) { //$NON-NLS-1$
	protected Boolean getValue() { return model.getInputConfig().isHeuristicForSampleBasedCriteria(); }
	protected void setValue(Object t) { model.getInputConfig().setHeuristicForSampleBasedCriteria((Boolean)t); }});
}
```

We can see that this exposes the option accessible via:

```java
model.getInputConfig().isHeuristicForSampleBasedCriteria()
```

So you have to jump into the function ```isHeuristicForSampleBasedCriteria```:

```java
public boolean isHeuristicForSampleBasedCriteria() {
	return config.isUseHeuristicSearchForSampleBasedCriteria();
}
```

And then you have to jump into ```isUseHeuristicSearchForSampleBasedCriteria```

```java
public boolean isUseHeuristicSearchForSampleBasedCriteria() {
        return heuristicSearchForSampleBasedCriteria;
}
```

And now you can take a look at the variable and see what its default is:

```java
private boolean heuristicSearchForSampleBasedCriteria = false;
```

-> The default should be "false".

idhamari: Thanks for the details. I missed the blue arrow, sorry. If you don't mind I will go for the first solution.

idhamari: Update: I modified [line 49](https://github.com/iaBIH/swtpreferences/blob/e5f2b398663f6e313973a3eb8f024d025633140d/src/main/de/linearbits/preferences/PreferenceBoolean.java#L49) in PreferenceBoolean.java and now the blue arrow is active

Screenshot from 2021-12-19 20-10-23

Next todo:
1. Push the updated swtpreferences code.
2. Correct the code in ARX i.e. replace true with false. 

Please let me know if you agree.

prasser: I don't understand what you mean by 2. Why "replace true with false". I think that the default parameter would need to be removed. Regarding 1. I would suggest  to not change line 49 but to add a new constructor ```public PreferenceBoolean(String label, Boolean hideLabel)```.

prasser: PS: Apart from that, I also think that it would be fine to keep the defaults in the ARX dialog. But they should be the correct defaults.

idhamari: >  I don't understand what you mean by 2. Why "replace true with false"? 


I meant lines 294, 334 and 340 in  e.g.

             window.addPreference(new PreferenceBoolean(Resources.getMessage("PropertyDialog.44"), true, true) { //$NON-NLS-1$
                protected Boolean getValue() { return model.getInputConfig().isHeuristicForSampleBasedCriteria(); }
                protected void setValue(Object t) { model.getInputConfig().setHeuristicForSampleBasedCriteria((Boolean)t); }});
            }

should be 

            window.addPreference(new PreferenceBoolean(Resources.getMessage("PropertyDialog.44"), false, true) { //$NON-NLS-1$
                protected Boolean getValue() { return model.getInputConfig().isHeuristicForSampleBasedCriteria(); }
                protected void setValue(Object t) { model.getInputConfig().setHeuristicForSampleBasedCriteria((Boolean)t); }});
            }


> I would suggest to not change line 49 but to add a new constructor public PreferenceBoolean(String label, Boolean hideLabel).

line 49 is modified to be:

        super(label,false);

I think this way the library reserves back-computability. I could not add a new constructor PreferenceBoolean(String label, Boolean hideLabel) as there is a constructor with the same signature i.e. 

 public PreferenceBoolean(String label, Boolean _default) 

> PS: Apart from that, I also think that it would be fine to keep the defaults in the ARX dialog. But they should be the correct defaults.

This is what I mean by "Correct the code in ARX"

idhamari:
The requested change is addressed in the last commit. Now all default values should be correct.

src/gui/org/deidentifier/arx/gui/view/impl/menu/DialogProperties.java
prasser: > line 49 is modified to be super(label,false);
> I think this way the library reserves back-computability.

No, this would break backwards compatibility (not "back-computability" ;)). The default is that if no default is provided, the default is "null". This would break this behaviour by setting the default to "false".

> I could not add a new constructor PreferenceBoolean(String label, Boolean hideLabel) as there is a constructor with the same signature i.e.

This is a really good point! So the solution would be to call
PreferenceBoolean(label, null, hideLabel) to achieve backward compatable behaviour.

-> No need to change swtpreferences.

prasser:
Thanks. Just to be sure: have you checked this for each variable for which a default is introduced by this PR?

idhamari:
> Thanks. Just to be sure: have you checked this for each variable for which a default is introduced by this PR?

Yes! there are only 3 of them.

Issue #381, opened 2022-04-11
[FEATURE] Add example for data-dependent DP


prasser:
The examples only contain a configuration with data-independent DP. Add variant with:

config.setHeuristicSearchStepLimit(steps); // Number of steps to search
config.setDPSearchBudget(dpSearchBudget) // Additional epsilon for search process

prasser:
@idhamari

PR #384, opened 2022-05-03
fix 381: add example for data-dependent DP

idhamari:
PR to fix [[FEATURE] Add example for data-dependent DP #381](https://github.com/arx-deidentifier/arx/issues/381)
Add a new example for  data-dependent DP!


prasser:
Issues with this PR:
- Please dont change existing code (parameter 0.00001d -> 0.999d). If you clearly need to change it, 0.1d is the highest value that could be accepted (and a better value than 0.00001 for the toy example).
- Mention the issue in the text with a "#" (e.g. "This PR addresses #381") so that it will be associated with the issue by GitHub.
- Setting searchSteps to 10000 doesnt make sense, because there cannot be more than 36 steps for the example dataset that is used here (product of height of all hierarchies: 3 * 2 * 6).
- Class is named "Example37_data_dependent_DP.java" but should just be "Example[NUMBER].java". So either you need to choose the next number for the example, or even better, add this to Examples37 with through methods: dataIndependentDP(...) and dataDependentDP(...)

idhamari:
> Class is named "Example37_data_dependent_DP.java" but should just be "Example[NUMBER].java". So either you need to choose the next number for the example, or even better, add this to Examples37 with through methods: dataIndependentDP(...) and dataDependentDP(...)

Done! the code is integrated in example37, and the new file is deleted. More clarification is added. 

> Please dont change existing code (parameter 0.00001d -> 0.999d). If you clearly need to change it, 0.1d is the highest value that could be accepted (and a better value than 0.00001 for the toy example).

Done! please note that for such a small dataset, all records are suppressed. 

> Mention the issue in the text with a "#" (e.g. "This PR addresses https://github.com/arx-deidentifier/arx/issues/381") so that it will be associated with the issue by GitHub.

Done! 

> Setting searchSteps to 10000 doesnt make sense, because there cannot be more than 36 steps for the example dataset that is used here (product of height of all hierarchies: 3 * 2 * 6).

Done!
done!

prasser:
> Done! Done! Done! Done! ...

No. ;)

Issue #365, opened 2021-11-26
Exclude suppressed records when analyzing the distribution of output data


srcds:
In “Analyze utility” > “Distribution” it would be valuable to have an option to exclude suppressed records when analyzing the distribution of the output dataset.

prasser:
Assigned to @srcds as this is a good oppportunity to better get to know ARX.

srcds:
Example:
1. Start ARX and open the "example.deid" project provided as an example in the "data" folder (https://github.com/arx-deidentifier/arx/blob/master/data/example.deid)
2. Anonymize the dataset (Edit -> Anonymize -> OK)
3. Navigate to the Analyze utility view (View -> Analyze Utility)
4. Click on the column of the "age" attribute 
5. Click on the Distribution tab in the lower part of the view

ARX should look at follows:
![Distribution_suppressed_records](https://user-images.githubusercontent.com/9593020/145117830-e3dec945-92a8-44be-96c2-0d6249b48c81.PNG)

During the anonymization some of the records were removed from the dataset. In the Distribution they are shown in a group labeled as "*" (Area 'A' in the figure). It would be nice to have an option to exclude all removed records from the distribution view. In another example it becomes apparent why:

![grafik](https://user-images.githubusercontent.com/9593020/145118324-bb19b672-b990-412f-93e3-5e4aa4d87a23.png)

Depending on whether the removed records are only removed from the distribution view or from the entire output data (open point for discussion), a button for toggling this new feature could be located in Area B or Area C.

idhamari:
The feature is added in this [commit ](https://github.com/iaBIH/arx/commit/43aea386ef48ef47b3d344980f3cb3d144f35f0c)

idhamari:
The suggested solution is in this [commmit](https://github.com/iaBIH/arx/commit/39ee30ece328bc59a1417f29b7795679f63b034b)

idhamari:
This issue is solved by this [PR](https://github.com/arx-deidentifier/arx/pull/391).

PR #388, opened 2022-06-13
Fix issue 365: Exclude suppressed records when analyzing the distribution of output data

idhamari:
Fix issue [Exclude suppressed records when analyzing the distribution of output data](https://github.com/arx-deidentifier/arx/issues/365)


prasser:
Won't merge. This PR implements the wrong approach. A parameter needs to be added to the classes/methods in package org.deidentifier.arx.aggregates which can be specified whether or not suppressed values should be included. This parameter can then be passed by the UI to obtain the according result.

PR #390, opened 2022-06-13
Overview on examples

tanjascats:
Created an overview of the existing examples of API usage for easier navigation through the examples. The descriptions are taken from the comments from the respective example files provided by their creators.


prasser:
Thanks, this is a great idea! Merging!

PR #391, opened 2022-06-13
Support showing / hiding special values (suppressed and null values)

prasser:
Support showing / hiding special values (suppressed and null values)


jenno-verdonck:
I would like to use this feature together with some new features in the master branch. Is it possible to get this PR merged or are there problems with?

Issue #392, opened 2022-06-14
[FEATURE] Overwrite warning for exporting hierarchies


srcds:
The buttons for importing and exporting hierarchies look similiar and are located next to each other. In addition, the file dialogues for loading and saving data look almost identical (at least in Windows). A user who **wants to import** a hierarchy can therefore **click export by accident**. This can lead to the user accidentally overwriting a hierarchy file irreversibly instead of loading its content. A warning asking whether a file should be overwritten could mitigate this risk.

Another mitigating design choice could be to place the "Create hierarchy" button between the ones for "Import hierarchy" and "Export hierarchy". However, i notices that this would break the categorization of the buttons in "File" and "Edit" functionalities.

prasser:
Assigning to srcds. This is almost trivial to implement. Give it a try :)

prasser:
The change introducing suggestions for filenames will point you to relevant places in the code.

PR #393, opened 2022-06-15
Implement #392 by enabling overwrite warning

srcds:


prasser:
Nice, clean and elegant.

PR #403, opened 2022-08-28
Add priority-based hierarchy builder and wizard

prasser:


ZachHaber:
@prasser There's something funky with the underlying hierarchy currently.

The hierarchy creation functionality works great and exactly as I'd like! Though a bit backwards. Highest to lowest is the default option, which goes in exact opposite order than what I'd expect. The mode of the data should be the last one to be dropped instead of the first with frequency prioritization (highest to lowest).

The issue I'm seeing is that the anonymizing doesn't apply the hierarchy as it displays in the "data transformation" tab.

On a dataset I'm using, (unfortunately I can't paste it for easy reference), the hierarchy when generating lowest to highest (to ensure that the mode of the data is the last dropped), looks like this:
image
If I generate the results, I see that it decides to drop the Race column instead of keeping "White" only, which is the result I got when I manually created this hierarchy earlier.

image

I can find the exact value that I used originally for this data set when I look for "Non-anonymous" transformations, which is really odd, because like I said, it should be the same result.
image

image


Instead of keeping "White", it kept "" (empty) as "Level-3" Race. I also tried the "Highest to Lowest" option, and the same result occurred.

When I edit the hierarchy and tell it to remove the underlying representation of the current hierarchy, it suddenly works exactly as it should, with the same results as v3.9.0 and the manual transformation settings I put together.
image

image
image

prasser:
Thanks a lot for trying this out so quickly!

We will take a look at the logic underlying the prioritization.

Just to be sure: have you anonymized the dataset again after changing the hierarchy? If yes, this looks like an additional bug unrelated to the new functionality. In this case, could you open a separate issue, please?

ZachHaber:
I've anonymized the dataset again after changing the hierarchy, including times where I use the "reset" button in between attempts to make sure that it's not my user-error!

ZachHaber:
To clarify, the issue I'm seeing with the hierarchy giving a different output than expected _is_ a bug in this PR, correct?

And the ordering of the hierarchy when setting it up from "Highest to Lowest" or "Lowest to Highest" is what you are asking me to open a different issue for?

prasser:
Ok, sounds like a separate bug then. Thanks a lot for diving into the details of ARX and reporting these types of problems. Highly appreciated! A separate issue for this would be great, as it's very likely not directly caused by the new functionality implemented in this branch.

prasser:
Additional question just to narrow this down: in the workflow your sketched above, an existing hierarchy is replaced with a new one, correct?

ZachHaber:
Correct, it was faster for me to test it by starting from an existing project, though the original "hierarchy" wasn't created through the hierarchy builder. I've also then cleared the data transformation table completely before re-creating the hierarchy and running the anonymization. I tend to try to be thorough for bug reports :)

prasser:
> I tend to try to be thorough for bug reports :)

I like that! It's a very important type of contribution! Thanks again!

> To clarify, the issue I'm seeing with the hierarchy giving a different output than expected is a bug in this PR, correct?

It's a bug, but likely not related to this PR but to the GUI code in general.

> And the ordering of the hierarchy when setting it up from "Highest to Lowest" or "Lowest to Highest" is what you are asking me to open a different issue for?

No. We will take a look at this, but the issue would be about the hierarch not being considered correctly, which is likely not related to this PR.

prasser:
> The hierarchy creation functionality works great and exactly as I'd like! Though a bit backwards. Highest to lowest is the default option, which goes in exact opposite order than what I'd expect. The mode of the data should be the last one to be dropped instead of the first with frequency prioritization (highest to lowest).

This has been fixed in commit e13899d

PR #418, opened 2022-10-25
remove hard coded links

idhamari:
To help with future updates. 

1. Remove hard-coded links
2. Use the same index as the sections 
3. All changes can be made only in one file i.e. messages.properties (e.g. restructuring the help folder tree in the help repository, adding new items, removing items, restructuring the items here, using a different server. One only needs to modify one file.


idhamari:
This PR:

- does not provide important improvement beside reducing the code.
-  may introduce bugs related to the ID changes.
- better solution is to move all help related to a separated file.

PR #420, opened 2022-11-02
Ia fix help modal text

idhamari:
The default value for modal help is false but the the box in the setting is checked


idhamari:
seems there were  some old testing commits, a new PR will be created!

PR #421, opened 2022-11-02
fix the label of modal help option

idhamari:
The default help modal option is false. The label should be corrected!


prasser:
No. The default is modal and everything is OK as it is.

idhamari:
My mistake, I mixed the modal vs non-modal definition.

PR #422, opened 2022-11-10
Fixed nullpointer on help

jenno-verdonck:
Opening the help menu caused a nullpointer as the model object is only available after a project is created. Will now by default open the help menu with modal on false. Another possible solution could be to disable the help menu when a project does not exist.


prasser:
Good catch. Thanks!

PR #423, opened 2022-11-14
Minor fix in HierarchyBuilderDate for Day of Week/Weekday

srcds:
Changed format for WEEKDAY (or “Day of Week” in GUI) from „u“ (year) to „e” (localized day-of-week). Reference:  https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html


prasser:
Is it really a "minor" fix? Sounds like a real bug.

srcds:
Well, the fix itself was minor. Anyway - apologies for the not optimal naming of the PR.

PR #429, opened 2022-12-01
Import or export all hierarchy files at once

idhamari:
This is useful when working with too many attributes, this PR imports and export all hierarchies from/to a folder at once. 

- The hierarchy filename must have the attribute name.  
- It must use the default separator and charset.


idhamari:
I guess seems more work needs to be done e.g. handling multiple files of the same attribute. 

I will close it for now until I add more improvement.

PR #430, opened 2022-12-02
view coding model slider value

idhamari:
Show the generalization and suppression factor value!


prasser:
Thx. If you want to display the value, please do it as a tool tip not a separate label.

idhamari:
Thanks for the feedback. 

> Please, also check whether the tool tip needs to be updated in other places as well, e.g. in public void setSelection(double selection).

I don't think this is needed, it works as expected. The tooltip is changed whenever the slider changes.

prasser:
Please externalize the string to messages.properties.

PR #431, opened 2022-12-12
show tooltip for attributes in weight view

idhamari:
This is useful when an attribute does not show up correctly on the label e.g. 
arxWeightsIssue


prasser:
Fixed by replacing Label with CLabel which provides the tooltip and a nicer text shortening method.

idhamari:
Cool!

PR #434, opened 2023-01-04
Arx saving hierarchies

idhamari:
Using same naming when save functional hierarchies!


prasser:
Won't merge. There are two issues with this PR:
(1) The way with which the name for the file is being generated is not safe. If interested, take a look at how this is done in the Controller class for the tabular hierarchies.
(2) By definition, functional hierarchies are independent of an actual variable and can be generic. While there might often be a direct association, the link to a specific variable is not as obvious as with tabular hierarchies.

Issue #433, opened 2022-12-15
[BUG] Inconsistent calculation metric using hierarchies vs hierarchyBuilders


jenno-verdonck:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**ARX GUI (please complete the following information):**
 - OS: [e.g. Windows]
 - Version [e.g. 3.7.1]

**Additional context**

prasser:
Thanks. This issue doesn't contain enough info to understand the potential bug. Please provide further details.

prasser:
PS: I'm pretty sure that this isn't a bug but intended behavior, but to be sure and to explain what is going on I need more details.

jenno-verdonck:
Yea my bad. I accidently posted the report already before finishing it.

prasser:
OK, thanks. As already suspected, this is not a bug but expected behaviour. In ARX, hierarchies that have been generated using the builders are assicated with a "functional definition" of the hierarchy as meta-information. This information can be used to more accurately measure information loss. One example:

Assume you have a dataset with an integer attribute. In the records, you have three values: 1, 3 and 7.

When using an interval-based hierarchy builder, you specify the interval [0, 10[. As a result, ARX knows that [0, 10[ is a generalization of 10 integer values and might, e.g., estimate information loss as 1/10 = 0.1

When loading a hierarchy from a CSV file, ARX cannot "understand" what the entries in the hierarchy mean. In the case of our example, it can just see that "[0, 10[" is a generalization of 1, 3 and 7 and might, e.g., estimate information loss as 1/|{1, 3, 7}| = 1/3 = 0.33

You can also save and load the functional definitions of hierarchies in the wizards, using the "Save..." and "Load..." buttons.

jenno-verdonck:
Thanks for the clarification.

I already suspected something like this. I can however see how this may be confusing for some users that expect the same result when visually seeing the same hierarchy in the GUI.

Calculating the score like it is done using the csv files seems to make more sense to me as it take into account the properties of the used dataset and more accurately reflects the score specific to the dataset. I suspect that therefor the utility of the dataset obtained using CSV files will be higher.

prasser:
> Calculating the score like it is done using the csv files seems to make more sense to me as it take into account the properties of the used dataset and more accurately reflects the score specific to the dataset. I suspect that therefor the utility of the dataset obtained using CSV files will be higher.

Not sure. I think this depends on the context and use case.

[...]

PR #436, opened 2023-01-04
loading .ahs files in the api

idhamari:
Handling [issue 433](https://github.com/arx-deidentifier/arx/issues/433). API currently does not provide a way to load .ahs files.  

One can test by modifying [example 24](https://github.com/arx-deidentifier/arx/blob/master/src/example/org/deidentifier/arx/examples/Example24.java) as follows: 

        builder1.save("age.ahs");
        HierarchyBuilderRedactionBased<?> ageBuilder = new HierarchyBuilderRedactionBased();
        ageBuilder.load("age.ahs", "age", data.getHandle());
        //data.getDefinition().setAttributeType("age", builder1);


prasser:
Not needed. You can use HierarchyBuilder.create(File file) to load ahs files.

idhamari:
I think after using create one still needs to prepare and build which main more coding when dealing with too many attributes. Probably the PR is not the best way to handle this but it saves time as one needs only one line per attribute.  

Still, I think it is better to keep this PR close until finding a better way.

Issue #437, opened 2023-01-05
[FEATURE] Release UI also for macs with M1 chip / ARM64


prasser:
macos-aarch64

prasser:
Thanks for the headsup. Adding this requires updating Eclipse/SWT to at least version 4.20, which was the first add aarch64 support for cocoa.

PR #438, opened 2023-01-05
Adds support for Mac M1/ Linux Arm64 - Fix issue #437 

danielporto:
Fixes issue #437 
swt dependencies downloaded from:
org.eclipse.swt.gtk.linux.aarch64: 
https://mvnrepository.com/artifact/org.eclipse.platform/org.eclipse.swt.gtk.linux.aarch64/3.116.100

org.eclipse.swt.cocoa.macosx.aarch64
https://search.maven.org/artifact/org.eclipse.platform/org.eclipse.swt.cocoa.macosx.aarch64/3.116.100/jar

Newer versions of these libs are available, but they were compiled with newer java jdk version than the one used in arx.


prasser:
Thanks a lot, this is great! Could you maybe:

(1) Replace SWT with version 14.100.v20200604-0951, because this is the version we currently use also for the other target platforms.
(2) Add those platforms to pom.xml as well (although it currently doesnt work).

?

prasser:
OK, I see. I think we also got them from Maven Central. It seems as if the version you took is the oldest version available of the M1 version of SWT. Let's try with those versions.

Issue #419, opened 2022-10-26
[BUG] Build failure > Compilation Error > CELL_TYPE_STRING not found


iflow:
**Describe the bug**

Building with Maven fails with various dependency errors, if using the suggested command of the README.

> [INFO] 4 errors
> [INFO] -------------------------------------------------------------
> [INFO] ------------------------------------------------------------------------
> [INFO] BUILD FAILURE
> [INFO] ------------------------------------------------------------------------
> [INFO] Total time:  39.249 s
> [INFO] Finished at: 2022-10-26T13:58:25+02:00
> [INFO] ------------------------------------------------------------------------
> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.6.1:compile (default-compile) on project arx: Compilation failure: Compilation failure:
> [ERROR] /D:/An/Projects/k-anonymity/arx/src/main/org/deidentifier/arx/io/ImportConfigurationExcel.java:[228,52] Symbol nicht gefunden
> [ERROR]   Symbol: Variable CELL_TYPE_STRING
> [ERROR]   Ort: Schnittstelle org.apache.poi.ss.usermodel.Cell
> [ERROR] /D:/An/Projects/k-anonymity/arx/src/main/org/deidentifier/arx/io/ImportAdapterExcel.java:[112,42] Symbol nicht gefunden
> [ERROR]   Symbol: Variable CREATE_NULL_AS_BLANK
> [ERROR]   Ort: Schnittstelle org.apache.poi.ss.usermodel.Row
> [ERROR] /D:/An/Projects/k-anonymity/arx/src/main/org/deidentifier/arx/io/ImportAdapterExcel.java:[188,53] Symbol nicht gefunden
> [ERROR]   Symbol: Variable CELL_TYPE_STRING
> [ERROR]   Ort: Schnittstelle org.apache.poi.ss.usermodel.Cell
> [ERROR] /D:/An/Projects/k-anonymity/arx/src/main/org/deidentifier/arx/io/ImportAdapterExcel.java:[255,82] Symbol nicht gefunden
> [ERROR]   Symbol: Variable CELL_TYPE_STRING
> [ERROR]   Ort: Schnittstelle org.apache.poi.ss.usermodel.Cell
> [ERROR] -> [Help 1]
> [ERROR]
> [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
> [ERROR] Re-run Maven using the -X switch to enable full debug logging.
[...]

PR #447, opened 2023-04-05
fix maven bug 419

idhamari:
fix [maven bug](https://github.com/arx-deidentifier/arx/issues/419)

Tested on both Windows 10 and Ubuntu 22.04

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,5 +16,7 @@ target
 .classpath
 .project
 /build
+/target
 .DS_Store
 /.pmd
+.vscode
--- a/pom.xml
+++ b/pom.xml
@@ -317,7 +317,7 @@
         <dependency>
             <groupId>com.github.ralfstuckert.pdfbox-layout</groupId>
             <artifactId>pdfbox2-layout</artifactId>
-            <version>1.0.0</version>
+            <version>1.0.1</version>
             <exclusions>
                 <exclusion>
                     <groupId>*</groupId>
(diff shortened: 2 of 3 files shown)


prasser:
Why habe you changed compile to install? I think that this change should be reverted.

idhamari:
> Why have you changed compile to install? I think that this change should be reverted.

As I understand, the goal is to use the library in a different project (probably another maven project). Using install will also add the library to maven local repository which saves time. If you wish, I can replace install with compile

prasser:
Yes, please.

prasser:
Downgrading POI from 4.1.1 to 3.10-FINAL is not an option due to the large number of vulnerabilities in this version. See here:

https://mvnrepository.com/artifact/org.apache.poi/poi

Maybe the other POI-related libraries also need to be updated. Or POI underwent an API-breaking change and ARX needs to be updated to work with never versions?

idhamari:
> Downgrading POI from 4.1.1 to 3.10-FINAL is not an option due to the large number of vulnerabilities in this version. 

I noticed 3.10libs are used in the [ant build here](https://github.com/arx-deidentifier/arx/tree/master/lib/ant/poi)

> ARX needs to be updated to work with newer versions?

I like this option.

prasser:
> I noticed 3.10libs are used in the [ant build here](https://github.com/arx-deidentifier/arx/tree/master/lib/ant/poi)

I know. Those should be updated as well.

>>ARX needs to be updated to work with newer versions?

> I like this option.

Then go for it! ;)

idhamari:
ok, I will do this during the weekend :)

prasser:
Needs to be revised

PR #468, opened 2025-09-19
Experimental mvn build

michaelsfr:
Added an experimental build script for Windows/Linux/Mac that sources local jar files for some libraries (poi, draw2d, smile and swt) instead of pulling from maven central (some versions are not available there).

Tested under Windows (11 and Git Bash) and Linux (Ubuntu 24.02).

### DIFF
--- a/install_deps.sh
+++ b/install_deps.sh
@@ -74,8 +74,3 @@ mvn install:install-file \
   -Dpackaging=jar
 
 echo "Dependencies installed successfully!"
-
-echo "Building with profile: $PROFILE"
-mvn clean package compile -P "$PROFILE" -Dcore=true -DskipTests
-
-echo "Build completed successfully!"
--- a/readme.md
+++ b/readme.md
@@ -33,9 +33,13 @@ Development setup
 
 Currently, the main development of ARX is carried out using Eclipse as an IDE and Ant as a build tool. Support for further IDEs such as IntelliJ IDEA and Maven is experimental.
 
-The Ant build script features various targets that can be used to build different versions of ARX (e.g. including GUI code or not). To build only the core code using Maven, set the system property `core` to `true`. This will build a platform independent jar with the ARX main code module and no GUI components:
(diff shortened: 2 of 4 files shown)


install_deps.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
+
+echo "Installing Maven dependencies..."
+
+mvn install:install-file \
```
prasser, line 5: There are many more dependencies that are needed by ARX and which are not on Maven central. Are you sure that this list is complete?

readme.md
```diff
@@ -33,9 +33,16 @@ Development setup
 
 Currently, the main development of ARX is carried out using Eclipse as an IDE and Ant as a build tool. Support for further IDEs such as IntelliJ IDEA and Maven is experimental.
 
-The Ant build script features various targets that can be used to build different versions of ARX (e.g. including GUI code or not). To build only the core code using Maven, set the system property `core` to `true`. This will build a platform independent jar with the ARX main code module and no GUI components:
+The Ant build script features various targets that can be used to build different versions of ARX (e.g. including GUI code or not).
 
-```$ mvn compile -Dcore=true``` 
+To build the project with Maven, run install script once to register dependencies, then execute the build process via Maven CLI to compile both the ARX main code module and GUI components:
```
prasser: run _the_ install script

prasser (commented):
See comments. BTW, this adresses #419

install_deps.sh
michaelsfr: All remaining dependencies are available on maven central or jitpack. I’ve just verified this again on both Linux and Windows using a clean local maven repository. These libs are the only ones that need to be installed manually first.

readme.md
michaelsfr: Done

install_deps.sh
prasser: You are right! I didn't know jitpack. Supercool! :)

